Issue #26, opened 2020-12-28
SDFormat8 Deprecated. Upgrade to SDFormat 9 or 10 recommended.


Alescontrela:
Installing sdformat8 with homebrew yields the following warning:

```
Warning: osrf/simulation/sdformat8 has been deprecated because it is past end-of-life date!
```

When possible, the repo and install instructions should be updated to make use of sdformat9/10.

PR #48, opened 2021-03-15
Upgrade to SDFormat10

### LINKED ISSUE #36: SDF 8 EOL?
I get

```
(base) Franks-Air:GTDynamics dellaert$ brew install sdformat8
==> Installing sdformat8 from osrf/simulation
Warning: osrf/simulation/sdformat8 has been deprecated because it is past end-of-life date!
```

varunagrawal:
This PR upgrades from the deprecated SDFormat8 to SDFormat9. Closes #36 and #26.

This upgrade had a bunch of breaking changes, but upgrading to SDFormat10 should not have any hassle. ~~I haven't used SDFormat10 since the docs still seem to be in `dev` mode rather than a final release.~~ SDFormat10 has an official release.


varunagrawal:
@dellaert can we merge this in soon? I know we discussed making the joint and link representations follow SDFormat better, but that is currently a tangential issue since we can upgrade SDFormat and then later continue on that. Merging this PR doesn't break anything, and since there's no movement happening on the representation update, this PR is sitting in limbo for over half a year.

Currently, this is blocking the CI since `brew install sdformat8` has an [issue on ubuntu 18.04](https://github.com/osrf/homebrew-simulation/issues/1200) and no debian package exists for sdformat8 (we have them for sdformat 6, 9, 10 and 11).

varunagrawal:
@dellaert I looked at the changes of this PR and the changes are consistent with the correct understanding of link and joint coordinate frames.

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -75,19 +75,18 @@ JointParams ParametersFromSdfJoint(const sdf::Joint &sdf_joint) {
 Pose3 GetJointFrame(const sdf::Joint &sdf_joint,
                     const LinkSharedPtr &parent_link,
                     const LinkSharedPtr &child_link) {
-  auto frame = sdf_joint.PoseFrame();
-  if (frame == "" || frame == child_link->name()) {
-    if (sdf_joint.Pose() == ignition::math::Pose3d())
+  auto frame = sdf_joint.PoseRelativeTo();
```
dellaert, line 78: Remove `auto` to show type, and change to `frame_name`

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -75,19 +75,18 @@ JointParams ParametersFromSdfJoint(const sdf::Joint &sdf_joint) {
...
+  Pose3 sdf_joint_pose = Pose3FromIgnition(sdf_joint.RawPose());
```
dellaert, line 79: This does not seem to be used. And I like that, because at this point there is no way to give it the standard `toTfrom` coordinate frame name. So, delete :-)

gtdynamics/universal_robot/sdf.cpp
dellaert: Can we document this function? I don't understand what it is supposed to do without reading the code, and it is not in the .h file (nor should it be!).

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -75,19 +75,18 @@ JointParams ParametersFromSdfJoint(const sdf::Joint &sdf_joint) {
...
+  if (frame.empty() || frame == child_link->name()) {
+    if (sdf_joint.RawPose() == ignition::math::Pose3d()) {
```
dellaert, line 81: Why even have this case? If it's identity we multiply with identity. So kill.

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -75,19 +75,18 @@ JointParams ParametersFromSdfJoint(const sdf::Joint &sdf_joint) {
...
       return child_link->wTl();
-    else
-      return child_link->wTl() * Pose3FromIgnition(sdf_joint.Pose());
+    } else {
+      return child_link->wTl() * Pose3FromIgnition(sdf_joint.RawPose());
```
dellaert, line 84: Can you
(a) document what this case means?
(b) assign `Pose3FromIgnition(sdf_joint.RawPose())` to the standard `T` conventions so I can check the indices match?

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -75,19 +75,18 @@ JointParams ParametersFromSdfJoint(const sdf::Joint &sdf_joint) {
...
+    }
   } else if (frame == parent_link->name()) {
-    if (sdf_joint.Pose() == ignition::math::Pose3d())
-      return parent_link->wTl();
-    else
-      return parent_link->wTl() * Pose3FromIgnition(sdf_joint.Pose());
+    return parent_link->wTl() * Pose3FromIgnition(sdf_joint.RawPose());
```
dellaert, line 87: same comment as above

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -75,19 +75,18 @@ JointParams ParametersFromSdfJoint(const sdf::Joint &sdf_joint) {
...
   } else if (frame == "world") {
-    return Pose3FromIgnition(sdf_joint.Pose());
+    return Pose3FromIgnition(sdf_joint.RawPose());
```
dellaert, line 89: same as above.

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -106,13 +105,34 @@ LinkSharedPtr LinkFromSdf(uint8_t id, const sdf::Link &sdf_link) {
   const auto &I = sdf_link.Inertial().Moi();
   inertia << I(0, 0), I(0, 1), I(0, 2), I(1, 0), I(1, 1), I(1, 2), I(2, 0),
       I(2, 1), I(2, 2);
-  const auto wTl = Pose3FromIgnition(sdf_link.Pose());
+
+  /// Call SemanticPose::Resolve so the pose is resolved to the correct frame
+  /// http://sdformat.org/tutorials?tut=pose_frame_semantics&ver=1.7&cat=specification&
+  // Get non-const Pose3d
```
dellaert, line 111: Document what pose is this, presumably:

``` 
// Get non-const pose of link in joint frame. 
```

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -106,13 +105,34 @@ LinkSharedPtr LinkFromSdf(uint8_t id, const sdf::Link &sdf_link) {
...
+  auto jTl = sdf_link.RawPose();
```
dellaert, line 112: maybe *not* use the naming convention now but `raw_pose` as this pose will be changed in place (very bad practice if you ask me).

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -106,13 +105,34 @@ LinkSharedPtr LinkFromSdf(uint8_t id, const sdf::Link &sdf_link) {
...
+  // Update from joint frame to base frame in-place.
+  // Base frame is denoted by "".
+  auto errors = sdf_link.SemanticPose().Resolve(jTl, "");
+  // If any errors in the resolution, throw an exception.
+  if (errors.size() > 0) {
+    throw std::runtime_error(errors[0].Message());
+  }
+  // Pose is updated from joint frame to world frame.
+  const auto wTl = Pose3FromIgnition(jTl);
```
dellaert, line 121: You say `w` but I say `b`, right? Or does "Pose3FromIgnition" magically change it from base frame to world frame?

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -106,13 +105,34 @@ LinkSharedPtr LinkFromSdf(uint8_t id, const sdf::Link &sdf_link) {
...
       I(2, 1), I(2, 2);
-  const auto wTl = Pose3FromIgnition(sdf_link.Pose());
+
+  /// Call SemanticPose::Resolve so the pose is resolved to the correct frame
+  /// http://sdformat.org/tutorials?tut=pose_frame_semantics&ver=1.7&cat=specification&
+  // Get non-const Pose3d
+  auto jTl = sdf_link.RawPose();
+  // Update from joint frame to base frame in-place.
+  // Base frame is denoted by "".
+  auto errors = sdf_link.SemanticPose().Resolve(jTl, "");
```
dellaert, line 115: I think the use of "Resolve" probably means we're not leaning into the new (typical/desired) SDF organization but I have no time to figure it out again.

dellaert (changes requested):
As a review again after changes :-)

gtdynamics/universal_robot/sdf.cpp
varunagrawal: This is documented in `sdf_internal.h`.

gtdynamics/universal_robot/sdf.cpp
varunagrawal: Yeah I found this strange too when I looked at this (it was pre-existing).

gtdynamics/universal_robot/sdf.cpp
varunagrawal: Agreed, but I am not going to get into a flamewar with the SDFormat folks on what style of programming to use.

gtdynamics/universal_robot/sdf.cpp
varunagrawal: It should be world frame throughout. The comment before is incorrect.

varunagrawal: Should be `// World frame is denoted by "".`

varunagrawal: Wait, I'm wrong. Let me look at this closer.

varunagrawal: Yeah it should be `bTl`. Nice catch.

gtdynamics/universal_robot/sdf.cpp
varunagrawal: Yep, pretty much. This would be a future change though, beyond the scope of this current PR and something that needs more effort (which I guess won't happen till the lab reconvenes in the fall).

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -75,19 +75,26 @@ JointParams ParametersFromSdfJoint(const sdf::Joint &sdf_joint) {
...
-      return parent_link->wTl() * Pose3FromIgnition(sdf_joint.Pose());
-  } else if (frame == "world") {
-    return Pose3FromIgnition(sdf_joint.Pose());
+  // Name of the coordinate frame the joint's pose is relative to.
+  // Specified by `relative_to` in the SDF file.
+  std::string frame_name = sdf_joint.PoseRelativeTo();
+
+  // Get the pose of the joint in the parent or child link's frame depending on
+  // the value of `frame_name`.
+  Pose3 lTj = Pose3FromIgnition(sdf_joint.RawPose());
```
dellaert, line 84: But, it's not lTj. We don't know that yet... Re-read my comments? :-)

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -106,11 +113,33 @@ LinkSharedPtr LinkFromSdf(uint8_t id, const sdf::Link &sdf_link) {
...
+    throw std::runtime_error(errors[0].Message());
+  }
+  // Pose is updated from joint frame to base frame.
+  const auto bTl = Pose3FromIgnition(raw_pose);
   const auto lTcom = Pose3FromIgnition(sdf_link.Inertial().Pose());
+
   return boost::make_shared<Link>(id, sdf_link.Name(),
                                   sdf_link.Inertial().MassMatrix().Mass(),
-                                  inertia, wTl, lTcom);
+                                  inertia, bTl, lTcom);
```
dellaert, line 135: So, here it says `bTl`, but the Link docs say `wTl`. I think it's fine, as I think Robot.h should not even know about the "world". But that's for a different PR. But, maybe you want to update the joint to also use the body frame above to be consistent within this file?

tests/testSdf.cpp
```diff
@@ -332,11 +332,11 @@ TEST(Joint, sdf_constructor_revolute) {
   EXPECT(assert_equal(screw_axis_j1_l1, j1->screwAxis(l1)));
 
   // Check transform from l0 com to l1 com at rest and at various angles.
-  Pose3 T_01comRest(Rot3::identity(), Point3(0, 0, 0.4));
+  Pose3 T_01com_rest(Rot3::identity(), Point3(0, 0, 0.4));
```
dellaert, line 335: rename all poses in this file to `toTfrom` ? E.g. `T_01com_rest` -> `com0Tcom1_rest`. I'm actually not sure about the comment in line 334. com0Tcom1_rest would be *from* com1 *to* com0, but the comment says different. If the comment *is* correct (?) should be `com1Tcom0_rest`

dellaert (changes requested):
Getting there.

gtdynamics/universal_robot/sdf.cpp
varunagrawal: The SDF spec mentions that all joint raw poses are with respect to the parent (or in some cases, the child) link's frame. This depends on what `relative_to` is set to for the joint. Here are the [docs](http://osrf-distributions.s3.amazonaws.com/sdformat/api/10.3.0/classsdf_1_1v10_1_1Joint.html#a69930bc6f744b43448db88d9255d4b1c).

gtdynamics/universal_robot/sdf.cpp
varunagrawal: This was what was confusing me in our call yesterday. Since Robot.h should not know about the world frame, everything should be in the body frame for the robot, is how I understood it, since we could then just update link's pose to be `wTb * bTl` since we have `bTl` for every link.

varunagrawal: Interestingly, the links are being parsed before the joints (since joints need to know about parent and child links). What this means is that, when parsing the links, we set the link's `wTl` to actually be `bTl` since the robot has no concept of the world frame (at least at the link level). Consequently, the joint's frame being set uses `link->wTl()` which is actually `link->bTl()` since we store `bTl` in the `wTl` property.

tests/testSdf.cpp
varunagrawal: #80 says we should name all the poses to `T_12`. I'll update the pose names here, but perhaps we should update that issue?

dellaert: Nope, you're right. We did agree on that rename and I forgot about it. Then just get rid of com? Still has to be `T_tf` in that order.

dellaert: And `M_tf` rather than `_rest`

gtdynamics/universal_robot/sdf.cpp
dellaert: I understand. So whether it is `lTj` is not known yet, hence my comment.

tests/testSdf.cpp
```diff
@@ -269,17 +269,17 @@ TEST(Joint, urdf_constructor_revolute) {
   EXPECT(j1->otherLink(l1) == l2);
 
   // rest transform
-  Pose3 T_12comRest(Rot3::Rx(0), Point3(0, 0, 2));
-  Pose3 T_21comRest(Rot3::Rx(0), Point3(0, 0, -2));
+  Pose3 com1Tcom2_rest(Rot3::Rx(0), Point3(0, 0, 2));
```
dellaert, line 272: Update as M_12 per my comments and #80

tests/testSdf.cpp
```diff
@@ -269,17 +269,17 @@ TEST(Joint, urdf_constructor_revolute) {
...
+  EXPECT(assert_equal(com2Tcom1_rest, j1->relativePoseOf(l1, 0.0)));
 
   // transform to (rotating -pi/2)
-  Pose3 T_12com(Rot3::Rx(-M_PI / 2), Point3(0, 1, 1));
-  Pose3 T_21com(Rot3::Rx(M_PI / 2), Point3(0, 1, -1));
-  EXPECT(assert_equal(T_12com, j1->relativePoseOf(l2, -M_PI / 2)));
-  EXPECT(assert_equal(T_21com, j1->relativePoseOf(l1, -M_PI / 2)));
+  Pose3 com1Tcom2(Rot3::Rx(-M_PI / 2), Point3(0, 1, 1));
+  Pose3 com2Tcom1(Rot3::Rx(M_PI / 2), Point3(0, 1, -1));
+  EXPECT(assert_equal(com1Tcom2, j1->relativePoseOf(l2, -M_PI / 2)));
```
dellaert, line 281: update to T_12

dellaert (approved):
Approve module those 3 comments

gtdynamics/universal_robot/sdf.cpp
varunagrawal: But it is known. It's known from the definition of the SDF file.

varunagrawal: Just need this last one resolved. Should I rename it to `raw_pose`?

varunagrawal:
Merging this. @dellaert we can discuss the last comment on `lTj` and address it in a future PR.

PR #139, opened 2021-04-08
Fixed Joint

varunagrawal:
Adding support for fixed joints so we can get the correct contact positions for non-trivial (aka non-spherical) links.

I am going through Lynch and Park so I don't get my screw theory in a twist (pun intended haha), so going to need a bit of hand-holding with respect to the computations.


varunagrawal:
PR is finally ready for review!

varunagrawal:
Ping!

gtdynamics/universal_robot/FixedJoint.h
```diff
@@ -0,0 +1,57 @@
...
+    return params;
+  }
+
+ public:
+  /**
+   * @brief Create FixedJoint using joint name, joint pose in
+   * base frame, and parent and child links.
+   *
+   * Since a fixed joint has 0 degrees of freedom, we set the screw axis to
+   * Vector6::Zero.
```
gchenfc, line 40: does this ever have ILS issues due to joint angle having no contribution?

python/tests/test_chain.py
```diff
@@ -120,7 +121,8 @@ def test_panda_FK(self):
             [0, -1, 0],
             [0, 0, -1]
         ])
-        expected_sM7 = Pose3(sR7, Point3(0.0882972, 0.00213401, 0.933844))
+        #regression
+        expected_sM7 = Pose3(sR7, Point3(0.098517, 0.004252, 0.971403))
```
gchenfc, line 125: can you clarify what's going on with the regressions in this file?  i.e. was it wrong before or did somehow the fixed joint change stuff?  or is it possible this is from a different PR and this is just a weird diff artifact?

python/tests/test_panda.py
```diff
@@ -45,7 +45,8 @@ def test_forward_kinematics(self):
             [0, -1, 0],
             [0, 0, -1]
         ])
-        expected_sT7 = Pose3(sR7, Point3(0.129315, 0.00227401, 0.88387))
+        # regression
+        expected_sT7 = Pose3(sR7, Point3(0.139535, 0.004392, 0.921429))
```
gchenfc, line 49: same comment about documenting regression reasoning (via reply to these comments)

gtdynamics/universal_robot/sdf.cpp
```diff
@@ -29,12 +30,21 @@ using gtsam::Pose3;
 sdf::Model GetSdf(const std::string &sdf_file_path,
                   const std::string &model_name,
                   const sdf::ParserConfig &config) {
-  sdf::SDFPtr sdf = sdf::readFile(sdf_file_path);
-  if (sdf == nullptr)
+  sdf::ParserConfig _config = config;
+  _config.URDFSetPreserveFixedJoint(true);
```
gchenfc, line 34: I was just thinking about a use case where someone (e.g. Jerred?  idk) doesn't care about the fixed joints and doesn't want the computation penalty of 70% more joints so they'd rather have the SDF do its auto-merging of links.  Could we maybe:
1) have a flag for whether to use the fixed joints or not
2) just remove these 2 lines and if you want to have explicit fixed joints then pass in the appropriate config?  (i.e. make erase-fixed-joints the default)

Because otherwise there's no way to un-preserve the fixed joints.

gchenfc (changes requested):
Cool

Though this seems like it will cause a decent amount of extra computations, e.g. for the A1 it looks like it's increasing the number of joints from 13 to 22 (if I counted correctly) so that's like 70% more computation, all consisting of expmap(0)-style computations?  Is this mainly for querying the correct points on bodies or are there actually kino-dynamics calculations that depend on preserving fixed joints?

I think in terms of convenience / less refactoring, this is good, and for people not using robot models containing fixed joints this doesn't have any drawbacks.  So the only main change I would like to see is the last review comment: allowing not preserving fixed joints as an option (e.g. by not overwriting the `config` argument or adding a flag).

python/tests/test_chain.py
varunagrawal: Links connected by fixed joints were getting merged into one link, which meant that the CoM was being changed to the weighted mean of the two. This is no longer the case.

gtdynamics/universal_robot/sdf.cpp
varunagrawal: Where is the computational penalty coming from? If it's forward kinematics then fixed joints don't participate in that since they don't have any values associated with them.
If it's product of exponentials, then it's just an extra matrix multiple for pTc and it should only be for linked connected to actuated links.

One argument is that we should be computing the extra matrix multiple since a particular chain configuration could involve the end effector being connected by a fixed link and a user wants to get the final fixed link rather than jump through mental hoops to compute them. In that perspective we're more in alignment with Lynch and Park.

In the case of the A1, we only have 4 extra links connected by fixed joints that will be actuated. The rest are all connected directly to the base and shouldn't be involved in any kino-dynamic computations. Moreover you can now easily compute sensor reference frames wrt the body frame which someone like Jerred would actually appreciate.

varunagrawal: I don't even see an argument for having this lumping of fixed links in the first place. This is only in URDF files which is why the flag doesn't exist for SDF files, and I haven't seen a single example where this lumping is useful.

Sure there is an extra computational cost in PoE but that's the cost we pay for making GTDynamics far more user friendly.

gtdynamics/universal_robot/FixedJoint.h
varunagrawal: It shouldn't but I'm not sure if a unit test actually tests for this.

gtdynamics/universal_robot/sdf.cpp
varunagrawal: Elaborating on the previous point, if all our models were SDF files only, then would the topic of computational complexity and lumping the fixed joints still be relevant?

My answer is no, so why should it be the case with URDF files? Why should we see different behaviors for each file format?

varunagrawal: Case in point: https://github.com/osrf/gazebo/issues/2732

My guess is that the lumping was just for the purpose of gazebo simulation. The rationale isn't documented anywhere so it's just hearsay at this point.

gchenfc: **Edit**: I just realized SdfConfig is only exposed in the `internal` namespace anyway so adding an option wouldn't really help, and propogating the option all the way up would be somewhat messy (exposing user to sdf:: doesn't seem the right direction).  Approving this PR, but I still think in the future there should be less computationally redundant ways to handle this.

---

> Where is the computational penalty coming from? If it's forward kinematics then fixed joints don't participate in that since they don't have any values associated with them. If it's product of exponentials, then it's just an extra matrix multiply for pTc and it should only be for links connected to actuated links.

I think fixed joints do have values associated with them, right?  They still have a joint angle key that is unconstrained and does nothing?  And all the corresponding factors too, so the factor graph increases in size by a decent bit?  Correct me if I'm wrong, but I'm fairly confident that this fixed joint will have all the same computations (including in "querying" and in factor graph solving) as a normal joint; it's just that the jacobians wrt joint angles will always be 0.

> The main concern is that you're thinking of what-ifs rather than what is necessary. If there is ever a use-case or need to have fixed joint lumping (which I highly doubt) then that person can create a test and make the appropriate change.

True, but I think I was thinking more in terms of _this_ is _changing_ the behavior to something less computationally efficient and it might be nice to leave the option in to _keep the existing_ more efficient behavior.  Really my main gripe is that these 2 lines make it impossible to not preserve fixed joints, when it would be just as reasonable to either make this behavior either a change-able default or a set-able non-default.

gchenfc (approved):

gtdynamics/universal_robot/sdf.cpp
varunagrawal: I'll add the flag and then merge this in.

PR #200, opened 2021-04-27
Feature/jr calibration

yetongumich:
Add the previous code to jointly perform camera calibration and calculating jumping robot configuration from images into the repo.

Note: the files maintained by Lucas have not been formatted/doscstrings by our standards, will make a later PR to clean it up.


examples/example_jumping_robot/tmp/camera_cal.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
```
varunagrawal, line 1: Is this file supposed to be here?

PR #234, opened 2021-08-11
Remove world language from joint

dellaert:
Removed two poses that refer to world frame in favor of two remaining poses, that we (Dan and I) renamed to "M", signifying relative poses at rest.

Next steps:
- remove wTl and wTcom altogether in Link.h
- Use formula (6) to calculate parentTchild(), that's a change from variable screwAxis to a constant screwAxis
- get rid of cScrewAxis_ and pScrewAxis_ as they will no longer be used

### DIFF
--- a/gtdynamics.i
+++ b/gtdynamics.i
@@ -213,9 +213,8 @@ class JointParams {
 
 virtual class Joint {
   uint8_t id() const;
-  const gtsam::Pose3 &wTj() const;
   const gtsam::Pose3 &jMp() const;
-  const Pose3 &jMc() const;
+  const gtsam::Pose3 &jMc() const;
   string name() const;
   gtdynamics::Link* otherLink(const gtdynamics::Link* link);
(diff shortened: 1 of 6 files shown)


yetongumich (approved):
LGTM! It should be fine to remove the world pose of the joint in the formulation, as it is never used for building the dynamics factor graph.

varunagrawal:
Can we merge this in after ICRA? I may have code that would break due to this change, and fixing regressions is something I don't want to do on a deadline.

dellaert:
Why would things break? I ran all unit tests. ICRA is a month away, so, not looking forward to that. Can you merge in unit tests that would break as a result of this?

dellaert:
BTW, @yetongumich and @gchenfc : if cable robot and/or jumping robot unit tests are not run in CI, please add a PR to enable them. Otherwise, face up to fixing those yourself if it turns out there are breaking changes :-)

PR #235, opened 2021-08-17
removed robot_ and updated accordingly

jerredchen:
Changes that I have made:
- Removed `robot_` attributes from Trajectory and updated affected code accordingly
- Updated affected code from removing the `Robot.isIdentical()` method
- Updated affected code from using `LinkSharedPtr` as opposed of `vector<string>` of link names in `Phase.h`

Issues to note:
- In `testStaticsSlice.cpp` the `Torque` method is producing different results but @varunagrawal said it is a bug that he already reported.
- In `testPhase.cpp` the resulting print statement from phase does not match the formatting of the unit test. The unit tests' resulting print is `tarsus_1_L1: [   0 0.19    0], tarsus_2_L2: [   0 0.19    0], tarsus_3_L3: [   0 0.19    0], ]` while the current phase print is:
```
[tarsus_1_L1:    0
0.19
   0, tarsus_2_L2:    0
0.19
   0, tarsus_3_L3:    0
0.19
   0, ]
```

### DIFF
--- a/gtdynamics/utils/Trajectory.cpp
+++ b/gtdynamics/utils/Trajectory.cpp
@@ -37,57 +37,58 @@ using std::vector;
 namespace gtdynamics {
 
 vector<NonlinearFactorGraph> Trajectory::getTransitionGraphs(
-    DynamicsGraph &graph_builder, double mu) const {
+    const Robot &robot, DynamicsGraph &graph_builder, double mu) const {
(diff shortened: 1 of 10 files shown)


dellaert (approved):
Awesome! I'll merge into Disha's branch. I think the unit tests was probably broken there already, so I'll comment on testPhase there.

PR #236, opened 2021-08-18
Cleanup & new ContactPointFactor

varunagrawal:
Added a new factor to allow for constraining between the link CoM and the point of contact.
Doing this rather than using a `NonlinearEqualityFactor` since the variable types are `Pose3` and `Point3`, and not a single type.

### DIFF
--- a/gtdynamics/factors/ContactPointFactor.h
+++ b/gtdynamics/factors/ContactPointFactor.h
@@ -1,0 +1,113 @@
+/* ----------------------------------------------------------------------------
+ * GTDynamics Copyright 2020, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * See LICENSE for the license information
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file  ContactPointFactor.h
+ * @brief Factor to enforce point on link is in contact with environment point.
+ * @author: Varun Agrawal
+ */
+
+#pragma once
+
+#include <gtsam/base/Matrix.h>
+#include <gtsam/base/Vector.h>
+#include <gtsam/geometry/Pose3.h>
+#include <gtsam/nonlinear/NonlinearFactor.h>
+
+#include <string>
+
+#include "gtdynamics/utils/ContactPoint.h"
+
+namespace gtdynamics {
+
+/**
+ * ContactPointFactor is a two-way nonlinear factor which enforces a
+ * point on the link to be equal to the point of contact in the environment.
+ */
+class ContactPointFactor
+    : public gtsam::NoiseModelFactor2<gtsam::Pose3, gtsam::Point3> {
+ private:
+  using This = ContactPointFactor;
+  using Base = gtsam::NoiseModelFactor2<gtsam::Pose3, gtsam::Point3>;
+
+  gtsam::Point3 lPc_;  // The contact point in the link's COM frame.
+
+ public:
+  /**
+   * Constructor.
+   *
+   * @param link_pose_key Key for the CoM pose of the link in contact.
+   * @param point_key Key for the contact point in the environment.
+   * @param cost_model Noise model associated with this factor.
+   * @param lPc Static transform from point of contact to link CoM.
+   */
+  ContactPointFactor(gtsam::Key link_pose_key, gtsam::Key point_key,
+                     const gtsam::noiseModel::Base::shared_ptr &cost_model,
+                     const gtsam::Point3 &lPc)
+      : Base(cost_model, link_pose_key, point_key), lPc_(lPc) {}
+
+  /**
+   * Convenience constructor which uses PointOnLink.
+   *
+   * @param point_on_link PointOnLink object which encapsulates the link and its
+   * contact point.
+   * @param point_key Key for the contact point in the environment.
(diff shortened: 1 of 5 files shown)


gtdynamics/factors/ContactPointFactor.h
```diff
@@ -0,0 +1,113 @@
...
+   * Constructor.
+   *
+   * @param link_pose_key Key for the CoM pose of the link in contact.
+   * @param point_key Key for the contact point in the environment.
+   * @param cost_model Noise model associated with this factor.
+   * @param lPc Static transform from point of contact to link CoM.
+   */
+  ContactPointFactor(gtsam::Key link_pose_key, gtsam::Key point_key,
+                     const gtsam::noiseModel::Base::shared_ptr &cost_model,
+                     const gtsam::Point3 &lPc)
```
dellaert, line 50: We have called this `point_com` consistently in, say, testPointGoalFactor.

dellaert (commented):
How is this different from PointGoalFactor? It could be different, just not very sure.

varunagrawal:
This is different because it is a binary factor that also estimates the point of contact. `PointGoalFactor` is a unary factor assuming we know the point of interest, or the point is a goal we wish the link to connect to.

gtdynamics/factors/ContactPointFactor.h
varunagrawal: `lPc` is also a point in the spatial frame that the link comes in contact with. Made sense to follow the GTSAM terminology.

dellaert:
> This is different because it is a binary factor that also estimates the point of contact. `PointGoalFactor` is a unary factor assuming we know the point of interest, or the point is a goal we wish the link to connect to.

Could you add this in the class comments?

gtdynamics/factors/ContactPointFactor.h
dellaert: I agree it makes sense, but it also makes the thing harder to understand. Looking at Daisha’s PR, we renamed it everywhere to “contact_in_com”. Could you please rename it for clarity?

gtdynamics/factors/ContactPointFactor.h
```diff
@@ -0,0 +1,113 @@
...
+
+#include <string>
+
+#include "gtdynamics/utils/ContactPoint.h"
+
+namespace gtdynamics {
+
+/**
+ * ContactPointFactor is a two-way nonlinear factor which enforces a
+ * point on the link to be equal to the point of contact in the environment.
```
dellaert, line 29: Explain better, as per your comment on my question.

gtdynamics/factors/ContactPointFactor.h
```diff
@@ -0,0 +1,113 @@
...
+   * @param point_on_link PointOnLink object which encapsulates the link and its
+   * contact point.
+   * @param point_key Key for the contact point in the environment.
+   * @param cost_model Noise model associated with this factor.
+   * @param t The time step at which to add the factor (default t=0).
+   */
+  ContactPointFactor(const PointOnLink &point_on_link, gtsam::Key point_key,
+                     const gtsam::noiseModel::Base::shared_ptr &cost_model,
+                     size_t t = 0)
+      : Base(cost_model,
```
dellaert, line 65: Delegate to constructor above?

gtdynamics/factors/ContactPointFactor.h
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Evaluate error.
+   * @param wTl The link's COM pose in the world frame.
+   * @param wPc The environment contact point in the world frame.
+   */
+  gtsam::Vector evaluateError(
+      const gtsam::Pose3 &wTl, const gtsam::Point3 &wPc,
+      boost::optional<gtsam::Matrix &> H_pose = boost::none,
+      boost::optional<gtsam::Matrix &> H_point = boost::none) const override {
+    gtsam::Vector error = gtsam::traits<gtsam::Point3>::Local(
```
dellaert, line 81: gtsam::traits<gtsam::Point3>::Local = minus. Just say

```
error = wPc - wTl.transformFrom(contact_in_com_, H_pose)
```

gtdynamics/factors/ForwardKinematicsFactor.h
```diff
@@ -53,7 +53,7 @@ class ForwardKinematicsFactor : public gtsam::BetweenFactor<gtsam::Pose3> {
    * @param end_link_name   The end link name whose pose to compute via FK.
    * @param joint_angles    gtsam::Values with joint angles for relevant joints.
    * @param model           The noise model for this factor.
-   * @param k               The integer time index
+   * @param k               The discretized time index.
```
dellaert, line 56: discrete time index

dellaert (changes requested):

gtdynamics/factors/ContactPointFactor.h
varunagrawal: I figured being consistent by using `traits::Local` would be great. That way no need to wonder which is the minuend and which is the subtrahend.

dellaert (approved):

dellaert:
nit: There is still a "discretized time index", change to "discrete time index" before merging.

PR #238, opened 2021-08-19
removing world context

danbarla:
changes included in first commit:

1. Link class refactored to hold one member : bMcom - link com in base frame at rest. wTl and lTcom removed. Constructor adjusted.
2. sdf reading functions adjusted. joints are defined in link frames (not COM) so we need to read the bTl in some places.
3. tests adjusted.

changes in second commit:

1. only change of naming in few files from wTj to bTj.

still missing:

1. need to adjust interface file (.i)
2. check python wrapper and tests
3. wTcom function still exists in Link class.

PR #240, opened 2021-08-21
Forward Kinematics Test with Pose Factors

varunagrawal:
Plus other niceties:

1. Delegated printing for various joint types.
2. Formatting.

@danbarla

### DIFF
--- a/gtdynamics/factors/PoseFactor.h
+++ b/gtdynamics/factors/PoseFactor.h
@@ -13,9 +13,6 @@
 
 #pragma once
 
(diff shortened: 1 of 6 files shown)


varunagrawal:
@dellaert this new test should ensure that when #238 lands, updating dependent code will be a cinch.

dellaert (approved):
Awesome!

varunagrawal:
Found some minor improvements I can make. Will update later today and then merge. :)

PR #241, opened 2021-08-23
updated README with new wrapper instructions

jerredchen:
README now has updated instructions on how to generate the GTDynamics Python bindings using the GTWRAP installed alongside GTSAM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,23 +107,30 @@ make
 
 GTDynamics now supports a Pybind11-based Python API.
 
-To start, please download and install the [GTwrap repository](https://github.com/borglab/wrap).
+GTWrap comes with the installation with GTSAM, which can also be used for generate the python bindings for GTDynamics.
 
 To compile and install the GTDynamics python library:
 
-1. In the build directory, run `cmake` with the flag `GTDYNAMICS_BUILD_PYTHON=ON`.
+1. Ensure that GTSAM is built with generated python bindings. If not, go to the build directory and run `cmake` with the flag `GTSAM_BUILD_PYTHON=ON`. Afterwards, install the GTSAM python package.
+
+    ```sh
+    cmake -DGTSAM_BUILD_PYTHON=ON ..
+    make && make python-install
+    ```
+
+2. In the GTDynamics build directory, run `cmake` with the flag `GTDYNAMICS_BUILD_PYTHON=ON`.
 
     ```sh
     cmake -DGTDYNAMICS_BUILD_PYTHON=ON ..
     ```
 
-2. Build as normal and install the python package.
+3. Build as normal and install the python package.
 
     ```sh
     make && make python-install
     ```
 
-3. To run the Python tests, you can simply run:
+4. To run the Python tests, you can simply run:
 
     ```sh
     make python-test


README.md
```diff
@@ -107,23 +107,30 @@ make
...
+GTWrap comes with the installation with GTSAM, which can also be used for generate the python bindings for GTDynamics.
 
 To compile and install the GTDynamics python library:
 
-1. In the build directory, run `cmake` with the flag `GTDYNAMICS_BUILD_PYTHON=ON`.
+1. Ensure that GTSAM is built with generated python bindings. If not, go to the build directory and run `cmake` with the flag `GTSAM_BUILD_PYTHON=ON`. Afterwards, install the GTSAM python package.
+
+    ```sh
+    cmake -DGTSAM_BUILD_PYTHON=ON ..
+    make && make python-install
```
varunagrawal, line 118: You can just link to the install instructions in GTSAM directly. That should be more helpful in an overall context.

dellaert: I agree a link would be nice, but does not preclude including these 2 lines for convenience.

README.md
```diff
@@ -107,23 +107,30 @@ make
...
+    ```
+
+2. In the GTDynamics build directory, run `cmake` with the flag `GTDYNAMICS_BUILD_PYTHON=ON`.
```
dellaert, line 121: I think you should explain that cmake in GTSAM and GTDynamics should ideally have the same, *non* /user/local prefix, and show the command with the corresponding prefix flag.

README.md
```diff
@@ -107,23 +107,30 @@ make
 
 GTDynamics now supports a Pybind11-based Python API.
 
-To start, please download and install the [GTwrap repository](https://github.com/borglab/wrap).
+GTWrap comes with the installation with GTSAM, which can also be used for generate the python bindings for GTDynamics.
```
varunagrawal, line 110: "comes bundled with GTSAM"

README.md
varunagrawal: Then the instructions should say `make && make install && make python-install`, since `make install` will actually install `gtwrap` for use.

README.md
varunagrawal: Also this should not be bullet point 2, since a user can choose to do a `apt install libgtsam` and install gtwrap separately, in which case this is point 1.

GTWrap instructions (which includes installation via GTSAM) and GTDynamics instructions should be separate subsections.

README.md
varunagrawal, line 110: This sentence does not make sense from an outsider's perspective. What can GTWrap "also be used" as an alternative to?

Perhaps do something like this?

```suggestion
#### GTWrap Prerequisite

To start, please download and install the [GTwrap repository](https://github.com/borglab/wrap).

GTWrap also comes bundled with GTSAM. Ensure that GTSAM is built with the generated python bindings. To do so, go to the build directory and run `cmake` with the flag `GTSAM_BUILD_PYTHON=ON`. Afterwards, install GTSAM along with its python package.

	```sh
	cmake -DGTSAM_BUILD_PYTHON=ON ..
	make && make install && make python-install
	```

#### GTDynamics Python Wrapper

>>> Add the rest of the instructions here.
```

dellaert: I specifically asked Jerred to *not* use a separate GTWrap implementation. If offered as an alternative, please create a totally different section, and let us not confuse the common use case.

README.md
jerredchen: @varunagrawal https://github.com/borglab/gtsam/blob/develop/python/README.md#readme from this readme it looks like that `make python-install` will install everything including `gtwrap`, but not sure if it's outdated.

PR #243, opened 2021-08-24
Python CI

varunagrawal:
This PR adds support for the Python CI, as well as adds some efficiency improvements.


varunagrawal:
@dellaert Gerry might not be awake so do you mind taking a look at this?

dellaert (approved):
Does it work and run the regular python tests? If so, paste a screenshot in PR comment and merge. If not, let's talk at 9?

varunagrawal:
Ye of such little faith

python-CI

PR #245, opened 2021-08-24
Remove kinpy as a dependency

varunagrawal:
Following from #242.


dellaert (changes requested):
Why are we doing this? kinpy is a great check on our kinematics, no?

varunagrawal:
Yes but we don't use it anywhere else. Having users import a package (e.g. on a robot) that doesn't provide core functionality seems wasteful.

Plus, I had kinpy print out the FK transforms and added them as regression values so that the test still does the same job. :)

dellaert (approved):
OK, thanks, that clears it up :-) !

dellaert:
Note, some checks seem to be failing...

PR #253, opened 2021-09-01
NAO URDF

varunagrawal:
Add customized version of NAO robot. Customization is to remove collisions between joint and link names so that the URDF can be correctly parsed.

@nrakoski3 to visualize the robot, you can checkout this branch and then run this:
```python
import gtdynamics as gtd
import roboplot  # Be sure to install package with `pip install -U roboplot`

robot = gtd.CreateRobotFromFile(gtd.URDF_PATH + "/nao.urdf")
roboplot.plot_robot(robot, "base_link")
```


varunagrawal:
For your viewing pleasure.

nao_plot

dellaert:
@nrakoski3 please review, @varunagrawal  made this for you…

nrakoski3 (commented):
When I run this file with:
robot = gtd.CreateRobotFromFile(gtd.URDF_PATH + "/humanoid/nao.urdf");
I receive the error:
RuntimeError: PoseRelativeToGraph error: multiple incoming edges to current vertex [LAnklePitch].

I think that the structure of the NAO URDF may not be compatible with the current implementation of CreateRobotFromFile

varunagrawal:
@nrakoski3 what version of SDFormat are you using?

varunagrawal:
`PoseRelativeToGraph` is an object from SDFormat so my guess is you're using a version of SDFormat below v10.

dellaert:
@nrakoski3 try it with docker image?

nrakoski3:
> `PoseRelativeToGraph` is an object from SDFormat so my guess is you're using a version of SDFormat below v10.

Hi Varun, I am using the current GT Dynamics Docker Image on master to run the code in a container. It has SDFormat10 installed. Please try running the code using that image to replicate my error.

Thank you,
Natalie Rakoski

varunagrawal:
Yup I did that yesterday, and it works for me. :(

varunagrawal:
Hey @nrakoski3, I emailed you a few weeks ago. Do you have time to figure this out soon?

varunagrawal:
ping @nrakoski3

gchenfc (approved):
Works for me!

2 notes:
1) It seems the version of roboplot I just installed from PyPI (1.0.3) needs to be updated to use `wTcom` instead of `bMcom` (or whatever the Link function changed to) - I think it appeared in like 5-6 places.
2) for posterity, it helps to put a `plt.show()` at the end of the example code Varun shared, so:
    ```python
    import gtdynamics as gtd
    import roboplot  # Be sure to install package with `pip install -U roboplot`
    import matplotlib.pyplot as plt
    
    robot = gtd.CreateRobotFromFile(gtd.URDF_PATH + "/nao.urdf")
    roboplot.plot_robot(robot, "base_link")
    plt.show()
    ```

By the way, thanks very much Varun!

varunagrawal:
Thanks for the quick review Gerry! Yeah I don't think I have updated `roboplot` for the latest API change so thanks for pointing that out.

Also, `plot_robot` takes a keyword parameter show, so you can just do `roboplot.plot_robot(robot, "base_link", show=True)`. The idea was that you could plot multiple things but only show on the last one, without the need to import matplotlib. :smile:

Issue #80, opened 2021-03-22
Big rename


dellaert:
- M_12 is for *rest* poses
- T_12 is for link poses (whereevr you choose to put them, on com or not)
- COM should be removed completely

varunagrawal:
@dellaert can we close this? What remains to completed for this?

dellaert:
Closed by #259

PR #259, opened 2021-09-02
Link class refactored to remove world context. 

dellaert:


dellaert:
@varunagrawal we merged to a branch not master. Can you approve so we can merge Dan's PR into master? @danbarla and I created Issue #260 to bring back wTl as bMlink.

varunagrawal (approved):
LGTM but we should keep an eye on resolving #80

varunagrawal:
In general it would be a good idea to link PRs to issues so we can track progress given we have a good amount of activity going on. :slightly_smiling_face:

dellaert:
> LGTM but we should keep an eye on resolving #80

Yeah, this PR goes towards #80

Issue #261, opened 2021-09-04
Memory corruption?


dellaert:
In PR #224, a test in testSpiderWalking gave inconsistent results. I spent quite some tracking it down to a subtle bug in initContactPointGoal, PR with fix coming.

PR #264, opened 2021-09-04
Detailed error check

dellaert:
Added machinery and a csv file to check individual factor error values, checking *all* 4,467 factor errors in testSpiderWalking by comparing them with stored values in the CSV file `tests/testSpiderWalking.csv`. This was useful to debug issue #261, but might also allow us to easily explore better initialization strategies (some of which already exist)


dellaert:
@yetongumich This PR exposes that here as well we have differences between Linux and Mac. If you click on a failing Linux check above you can see that the errors are very different on Linux from the one in the CSV file, on my mac. If you are running this on Linux, could you do a valgrind? Asking you as @varunagrawal and @gchenfc are gunning for ICRA...

dellaert:
@danbarla also feel free to run valgrind if you're on Linux. Want to get to the bottom of this - might point to (another) weird bug.

danbarla (commented):
[spider_walking_valgrind_log_save.txt](https://github.com/borglab/GTDynamics/files/7112300/spider_walking_valgrind_log_save.txt)

This is the log of my valgrind run on my ubuntu 20 pc.
some parts on the sdf.cpp look suspicious (potentially my fault :) )
But I'm not sure these are the cause of this anomaly (which I see also on my machine).

dellaert:
@danbarla I just checked in some regressions on initial values, with expected values from my Mac, and the joint angles and joint velocities do not agree on Linux (used the docker container :-))
```
/usr/src/GTDynamics/tests/testSpiderWalking.cpp:204: Failure: "expected -1.8499999999999999e-05 but was: 6.0276815148976885e-06" 
/usr/src/GTDynamics/tests/testSpiderWalking.cpp:205: Failure: "expected 1.9759999999999998e-06 but was: -5.1303190182304889e-07" 
/usr/src/GTDynamics/tests/testSpiderWalking.cpp:206: Failure: "expected -1.8499999999999999e-05 but was: 6.0276815148976885e-06" 
/usr/src/GTDynamics/tests/testSpiderWalking.cpp:207: Failure: "expected 1.9759999999999998e-06 but was: -5.1303190182304889e-07" 
/usr/src/GTDynamics/tests/testSpiderWalking.cpp:209: Failure: "expected -3.8999999999999999e-06 but was: 3.7040939584755143e-06" 
/usr/src/GTDynamics/tests/testSpiderWalking.cpp:210: Failure: "expected -5.1100000000000002e-06 but was: 2.597843754075192e-06" 
/usr/src/GTDynamics/tests/testSpiderWalking.cpp:211: Failure: "expected -3.8999999999999999e-06 but was: 3.7040939584755143e-06" 
/usr/src/GTDynamics/tests/testSpiderWalking.cpp:212: Failure: "expected -5.1100000000000002e-06 but was: 2.597843754075192e-06" 
```
That confirms my suspicion, and narrows the problem down to the initialization of joint angles/velocities, as Poses seem to agree.

tests/testSpiderWalking.cpp
```diff
@@ -142,11 +143,73 @@ TEST(testSpiderWalking, WholeEnchilada) {
...
 
+  // Compare error for all factors with expected values in file.
+  // Note, expects space after comma in csv or won't work.
+  std::string key, comm;
+  double expected;
+  std::string filename = kTestPath + std::string("/testSpiderWalking.csv");
+  std::ifstream is(filename.c_str());
+  is >> key >> expected;
+  double actual = graph.error(init_vals);
+  const double tol = 1.0;
```
yetongumich, line 154: Is 1.0 too large for tolerance?

tests/testSpiderWalking.cpp
```diff
@@ -142,11 +143,73 @@ TEST(testSpiderWalking, WholeEnchilada) {
...
+  EXPECT_DOUBLES_EQUAL(expected, actual, tol);
+
+  // If there is an error, create a file errors.csv with all error comparisons.
+  if (fabs(actual - expected) > tol) {
```
yetongumich, line 158: This functionality is very helpful! Maybe we should make this a general function which everyone can use?

BTW, I used to write some similar functions for debugging. I hope we had this earlier.

yetongumich (approved):
LGTM with minor comments. I think the functionality is very useful. Maybe we should wrap it as a function/in a class that everyone can use.

danbarla:
Please have a look at this link:  
[https://github.com/imneme/pcg-cpp/issues/50]

there could be a problem with std::normal_distribution when working cross_platform. The order of the sampled values is changed.
actually every pair of values sampled is the same but in the opposite order.

I was able to prove this by adding this line at Sampler.cpp line 43 in GTSAM and recompiling:
`      Normal dist(0.0, sigma);
      auto res = dist(generator_);
      result(i) = dist(generator_);`

after this the test passes except line 183 which I think will pass also with the correct init_vals

dellaert:
> @dellaert
> Please have a look at this link:
> [https://github.com/[imneme/pcg-cpp/issues/50](https://github.com/imneme/pcg-cpp/issues/50)]
> 
> there could be a problem with std::normal_distribution when working cross_platform. The order of the sampled values is changed.
> actually every pair of values sampled is the same but in the opposite order. This happens even though we use the same seed.
> 
> I was able to prove this by adding this line at Sampler.cpp line 43 in GTSAM and recompiling:
> ` Normal dist(0.0, sigma); auto res = dist(generator_); result(i) = dist(generator_);`
> 
> after this the test passes except line 183 (still not sure why)

Woah, good find!!! That is very annoying behavior :-( 
Can you suggest a way to fix it? Or several options?

dellaert:
Actually, we don’t have to fix it: randomness in unit tests is frowned upon, anyway, and is actually not needed here. Rather, can we change the test (and possibly the init code) so we can ask for a sequence with no randomness ?

gtdynamics/utils/Phase.cpp
```diff
@@ -67,6 +67,8 @@ Phase::ContactPointGoals Phase::updateContactPointGoals(
     const string &name = kv.first;
     const Point3 &cp_goal = cp_goals.at(name);
     const bool stance = hasContact(name);
+    // If a contact is not on a stance leg, it is on a swing leg and we advance
+    // the contac goal by adding the 3-vector `step`.
```
varunagrawal, line 71: typo "contact" not "contac"

tests/testSpiderWalking.cpp
```diff
@@ -142,11 +143,73 @@ TEST(testSpiderWalking, WholeEnchilada) {
...
   auto results = optimizer.optimize();
 
-  // TODO(frank): test whether it works
+  // Regression!
+  EXPECT_DOUBLES_EQUAL(986936294413055, graph.error(init_vals), 100);
+  EXPECT_DOUBLES_EQUAL(353211972620861, graph.error(results), 100);
+
+  // Add regressions on initial values.
+  auto body = robot.link("body");
+  EXPECT(gtsam::assert_equal(Pose3(Rot3(), Point3(0, 0, 0)),
```
varunagrawal, line 187: Replace with `Pose3()`

tests/testSpiderWalking.cpp
```diff
@@ -142,11 +143,73 @@ TEST(testSpiderWalking, WholeEnchilada) {
...
+                             Pose(init_vals, body->id(), 0), 1e-3));
+
+  // TODO(frank): why is this still at the identity?
+  EXPECT(gtsam::assert_equal(Pose3(Rot3(), Point3(0, 0, 0)),
```
varunagrawal, line 191: Same as above. Though the fact that it is identity at `t=K` is concerning.

danbarla:
1. we could set the gaussian noise to zero. this will just initialize all values to exactly zero.  
2. We could add a boost::optional argument with external initialization to `MultiPhaseZeroValuesTrajectory ` which will come from a csv file. This can help in testing but it will require writing new code and the real code that runs when there is no external initializaion will not be tested.
3. We can just not test for specific values here. Is this test different from other tests where we initialize with noise and not check for specific values?

tests/testSpiderWalking.cpp
dellaert: Different PR :-) Feel free after this lands to generalize *iff* you use it somewhere else.

tests/testSpiderWalking.cpp
dellaert: No, switched to rtol anyway

dellaert:
@danbarla I took your zero noise suggestion, implemented it, and merged. I also merged the branch depending on this. Next (after teaching) I will push the branch you and I prototyped in response to the issue I created.

PR #280, opened 2021-09-22
Joint Measurement Factor

varunagrawal:
This PR adds a new `LinkPoseFactor` to relate the poses of the parent and child links of a joint given the joint coordinate as a measurement.

This is in contrast to the `PoseFactor` which takes the joint coordinate as a value.

### DIFF
--- a/gtdynamics.i
+++ b/gtdynamics.i
@@ -15,17 +15,24 @@ const string SDF_PATH = kSdfPath;
 const gtsam::KeyFormatter GTDKeyFormatter;
 
 /********************** factors **********************/
-#include <gtdynamics/factors/LinkPoseFactor.h>
+#include <gtdynamics/factors/JointMeasurementFactor.h>
 template <JOINT>
-class LinkPoseFactor : gtsam::NonlinearFactor {
-  LinkPoseFactor(gtsam::Key wTp_key, gtsam::Key wTc_key,
-                 const gtsam::noiseModel::Base *cost_model,
-                 const gtdynamics::Joint *joint,
-                 const JOINT::JointCoordinate &joint_coordinate, size_t k);
+class JointMeasurementFactor : gtsam::NonlinearFactor {
+  JointMeasurementFactor(gtsam::Key wTp_key, gtsam::Key wTc_key,
+                         const gtsam::noiseModel::Base *cost_model,
+                         const gtdynamics::Joint *joint,
+                         const JOINT::JointCoordinate &joint_coordinate,
+                         size_t k);
+  JointMeasurementFactor(const gtsam::noiseModel::Base::shared_ptr &model,
+                         const gtdynamics::Joint *joint,
+                         const JOINT::JointCoordinate &joint_coordinate,
+                         size_t k);
 
   void print(const string &s = "", const gtsam::KeyFormatter &keyFormatter =
                                        gtdynamics::GTDKeyFormatter);
 };
+
+typedef gtdynamics::JointMeasurementFactor<gtdynamics::RevoluteJoint> RevoluteJointMeasurementFactor;
 
 #include <gtdynamics/factors/PoseFactor.h>
 class PoseFactor : gtsam::NonlinearFactor {
--- a/gtdynamics.i
+++ b/gtdynamics.i
@@ -42,6 +42,8 @@ class PoseFactor : gtsam::NonlinearFactor {
 
   void print(const string &s="",
              const gtsam::KeyFormatter &keyFormatter=gtdynamics::GTDKeyFormatter);
+
+  gtsam::Vector unwhitenedError(const gtsam::Values& x) const;
 };
 
 #include <gtdynamics/factors/ForwardKinematicsFactor.h>
@@ -62,6 +64,8 @@ class ForwardKinematicsFactor : gtsam::NoiseModelFactor {
   void print(const string &s="",
              const gtsam::KeyFormatter &keyFormatter=gtdynamics::GTDKeyFormatter);
   const gtsam::Pose3 measured() const;
+
+  gtsam::Vector evaluateError(const gtsam::Pose3& p1, const gtsam::Pose3& p2) const;
 };
 
 #include <gtdynamics/factors/ContactEqualityFactor.h>
@@ -191,6 +195,8 @@ class ContactHeightFactor : gtsam::NonlinearFactor {
 
   void print(const string &s = "", const gtsam::KeyFormatter &keyFormatter =
                                        gtdynamics::GTDKeyFormatter);
+
+  gtsam::Vector evaluateError(const gtsam::Pose3 &sTl) const;
 };
 
 /********************** link **********************/
--- a/gtdynamics.i
+++ b/gtdynamics.i
@@ -15,6 +15,18 @@ const string SDF_PATH = kSdfPath;
 const gtsam::KeyFormatter GTDKeyFormatter;
 
 /********************** factors **********************/
+#include <gtdynamics/factors/LinkPoseFactor.h>
+template <JOINT>
+class LinkPoseFactor : gtsam::NonlinearFactor {
+  LinkPoseFactor(gtsam::Key wTp_key, gtsam::Key wTc_key,
+                 const gtsam::noiseModel::Base *cost_model,
+                 const gtdynamics::Joint *joint,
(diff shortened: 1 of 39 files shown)


gtdynamics.i
```diff
@@ -15,6 +15,18 @@ const string SDF_PATH = kSdfPath;
 const gtsam::KeyFormatter GTDKeyFormatter;
 
 /********************** factors **********************/
+#include <gtdynamics/factors/LinkPoseFactor.h>
+template <JOINT>
```
gchenfc, line 19: I haven't tested this but doesn't this need to have
`template <JOINT=gtdynamics::RevoluteJoint>` or something?

Maybe if you could make a short python unit test just to check it can import properly?

gtdynamics/factors/LinkPoseFactor.h
```diff
@@ -0,0 +1,109 @@
...
+   * @param model The noise model for this factor.
+   * @param joint The joint connecting the links.
+   * @param joint_coordinate The coordinates of the joint motion.
+   * @param k The time index.
+   */
+  LinkPoseFactor(gtsam::Key wTp_key, gtsam::Key wTc_key,
+                 const gtsam::noiseModel::Base::shared_ptr& model,
+                 const JointConstSharedPtr joint,
+                 const typename JOINT::JointCoordinate& joint_coordinate,
+                 size_t k)
```
gchenfc, line 55: `PoseFactor` avoids needing to pass in `k` by using `wTp_key.time()`.  That might help avoid situations where someone accidentally passes the wrong `k` and then gets a runtime error.

gtdynamics/factors/LinkPoseFactor.h
```diff
@@ -0,0 +1,109 @@
...
+      const gtsam::Pose3& wTp, const gtsam::Pose3& wTc,
+      boost::optional<gtsam::Matrix&> H_wTp = boost::none,
+      boost::optional<gtsam::Matrix&> H_wTc = boost::none) const override {
+    gtsam::Values joint_angles;
+    InsertJointAngle(&joint_angles, joint_->id(), k_, joint_coordinate_);
+
+    gtsam::Matrix H_q;
+    gtsam::Matrix6 H;
+    gtsam::Pose3 wTc_hat =
+        joint_->poseOf(joint_->child(), wTp, joint_angles, k_, H_wTp, H_q);
```
gchenfc, line 89: Can you omit `H_q` or pass `boost::none`?

gtdynamics/factors/LinkPoseFactor.h
```diff
@@ -0,0 +1,109 @@
...
+namespace gtdynamics {
+
+/**
+ * @brief A 2-way factor to relate the parent and child links of a joint given
+ * the joint coordinate as a measurement.
+ *
+ * @tparam JOINT The type of joint which is being constrained.
+ */
+template <typename JOINT>
+class LinkPoseFactor
```
gchenfc, line 30: Not a big fan of this name since it doesn't really describe what the difference between this and `PoseFactor` is.

I appreciate that it's difficult to come up with a name, and that not liking it without providing an alternative is annoying.  Maybe `PoseFactorMeasurement` or `NumericalPoseFactor` or `PoseFactorWithAngle` something?  Ehh not sure, but I'm not a huge fan of LinkPoseFactor.

PS I'm not opposed to renaming both to `LinkPoseFactor`, but atm I don't see how `Link` describes the difference between these two factors.

gtdynamics/factors/LinkPoseFactor.h
```diff
@@ -0,0 +1,109 @@
...
+             const gtsam::KeyFormatter& keyFormatter =
+                 gtsam::DefaultKeyFormatter) const override {
+    std::cout << s << "LinkPoseFactor(" << keyFormatter(key1()) << ","
+              << keyFormatter(key2()) << ")\n";
+    gtsam::traits<typename JOINT::JointCoordinate>::Print(joint_coordinate_,
+                                                          "  measured: ");
+    this->noiseModel_->print("  noise model: ");
+  }
+};
+}  // namespace gtdynamics
```
gchenfc, line 109: Newline?

gtdynamics.i
```diff
@@ -15,6 +15,18 @@ const string SDF_PATH = kSdfPath;
...
+class LinkPoseFactor : gtsam::NonlinearFactor {
+  LinkPoseFactor(gtsam::Key wTp_key, gtsam::Key wTc_key,
+                 const gtsam::noiseModel::Base *cost_model,
+                 const gtdynamics::Joint *joint,
+                 const JOINT::JointCoordinate &joint_coordinate, size_t k);
+
```
gchenfc, line 25: add second constructor too?  (optional)

gchenfc (changes requested):
Looks good - mainly soft request to add python unit test (or remove changes from gtdynamics.i and let someone else add when they need it) and the name `LinkPoseFactor` is not super great IMO.

Otherwise just nits.

gtdynamics/factors/LinkPoseFactor.h
varunagrawal: Yeah that's a good point. I have not been a fan of the `PoseFactor` name either but didn't know what would be a good way to name this and convey intent. Maybe `JointCoordinateFactor` or `JointMeasurementFactor`?

gtdynamics/factors/LinkPoseFactor.h
varunagrawal: Ah this was leftover from when I was debugging something, nice catch.

gtdynamics/factors/LinkPoseFactor.h
varunagrawal: Huh then I've been doing this all wrong so far. I should have been using this idiom all along.

gtdynamics.i
varunagrawal: Oooh good call. I think I may have discovered a bug in the wrapper thanks to this.

gtdynamics/factors/LinkPoseFactor.h
gchenfc: Personally I would vote for `JointMeasurementFactor`.

Although it kind of sounds like a prior factor on a joint angle variable, but since that's more like FKprior or just PriorFactor then I think it's sufficiently unambiguous.

varunagrawal: It's not a prior right? It's a constraint on the link poses based on the joint measurement.

gtdynamics/factors/LinkPoseFactor.h
varunagrawal: OKay there's a flaw in this approach in that I should be able to pass in a `gtsam::Key`. For example, a lot of my work with the IMU factors involves using a `gtsam::Key` to refer to the base link of a robot.

varunagrawal:
The tests will pass once we merge the pending PR in `gtwrap`.

gtdynamics/factors/LinkPoseFactor.h
gchenfc: Right - so my wording as confusing but what I was trying to say is that it sounds as if `JointMeasurementFactor` is a prior factor, but it is _not_ a prior factor: therefore it could be confusing.  Specifically, `JointMeasurementFactor` sounds like it's a prior factor on the joint angle: `gtsam::PriorFactor<double>(JointAngleKey(id, t), measurement)`

However, I think that because there's more obvious names for prior factors (e.g. FKprior, or just PriorFactor), then it's sufficiently clear that `JointMeasurementFactor` is not a prior factor.

Anyway, we're on the same page so we're good.

gtdynamics/factors/LinkPoseFactor.h
gchenfc: gtd::DynamicsSymbol can be constructed from a key, so for example could do

`LinkPoseFactor(gtd::DynamicsSymbol(key1), gtd::DynamicsSymbol(key2), model, joint, 0.1)`

Or could overload the constructor

If `key1` and `key2` don't have semantically meaningful time indices to begin with then it kind of gets messed up anyway to connect a properly timed joint to a non-timed link?

I think it might be good practice for all links to be referred to by DynamicsSymbols rather than raw keys?

gchenfc (approved):
EDIT: Forgot about the timestamp `k` issue in the constructor - would like to resolve between us before merge.


LGTM!

(pending gtwap/CI)

PR #284, opened 2021-09-27
Remove need for leading slash when loading robot models

varunagrawal:
The need for the leading slash (e.g. `/tests/simple_rr.sdf`) is counter-intuitive and error-prone.
This PR fixes that.


dellaert:
Why ask my review here?

varunagrawal:
To move things along a bit faster

gchenfc (approved):

PR #285, opened 2021-10-03
Add KinematicsParameters to python wrapper

gchenfc:
Title self-explanatory.

### DIFF
--- a/gtdynamics.i
+++ b/gtdynamics.i
@@ -354,6 +354,13 @@ class PointOnLink {
 
 // PointOnLinks defined in specializations.h
 
+/********************** Optimizer **********************/
+#include <gtdynamics/optimizer/Optimizer.h>
+class OptimizationParameters {
+  gtsam::LevenbergMarquardtParams lm_parameters;
+  OptimizationParameters();
+};
+
 /********************** kinematics **********************/
 #include <gtdynamics/kinematics/Kinematics.h>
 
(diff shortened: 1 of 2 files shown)


varunagrawal (approved):
There is special syntax for a pointer. It's the ampersand symbol, can you try with that? It should be cleaner.

gchenfc:
I'm familiar with the @, *, and &, but where are you proposing to use the pointer?   (btw, I think you maybe meant "at" symbol, ampersand is normal reference right?)

This is a const-reference to a boost::shared pointer of a const - so I'm not sure how to make this cleaner?

We can delete the const ref - I guess the const ref saves one counter increment/decrement that won't (?) get compiler optimized, but I doubt that makes much of a difference.

dellaert:
KISS? * in a .i files already means “shared pointer”. Usage in point should be:
```
p = Params(…)
Some_functionthat_takes_shared_pointer(p) 
```
and in .i file:
```
class Params{ …};
void Some_functionthat_takes_shared_pointer(Params* p);
```
and clean up ref/const in c++ to match:
```
void Some_functionthat_takes_shared_pointer(const boost::shared_ptr<Params>& p);
```

gchenfc:
Wow thanks I'm stupid, I don't know why I thought that wouldn't work but in retrospect it should have been obvious.  Thanks!

PR #289, opened 2021-10-11
Add models from the DARPA SubT challenge

### LINKED ISSUE #281: All your robots belong to us!
Can we get the legged sdfs from the subT challenge to load?
https://github.com/osrf/subt/wiki/Robots

varunagrawal:
Fixes #281.


varunagrawal:
ping @gchenfc

gchenfc:
Sorry I must have missed this the first time around - I think the target branch is wrong?  Or if you're intentionally trying to keep a separate branch, maybe you can create a separate PR first to merge master (or just perform locally) before this PR so it's easier to read?

gchenfc:
Oh nevermind, but how come the "files changed" diffs are all messed up on github?

gchenfc (approved):
Just reviewing based on commit 6e1de74f6742 which, correct me if I'm wrong, is the only real commit relevant to this PR.

This is fine, but perhaps in the future we should consider also reorganizing the urdfs / sdfs into robot-oriented directories (e.g. subt / spider / a1 / cablerobot / etc) instead of sdf / urdf / subt.  Since loading syntax doesn't change depending on the file type, and it doesn't really make sense to have folders for file types and a folder that's categorically named that contains sdfs.  Or maybe these could go in a `models/sdf/subt` folder?  Also just realized, why are most of these files except `cerberus...` in a `subt/subT` subfolder?

I know sdf is sensitive to folder structure so maybe some of these suggestions wouldn't work.

varunagrawal:
Yeah that's because I merged in master and Github doesn't seem to like that. It should resolve itself once we merge in `robot-models`.

varunagrawal:
I guess I can remove the inner subT folder.

varunagrawal:
> This is fine, but perhaps in the future we should consider also reorganizing the urdfs / sdfs into robot-oriented directories (e.g. subt / spider / a1 / cablerobot / etc) instead of sdf / urdf / subt. Since loading syntax doesn't change depending on the file type, and it doesn't really make sense to have folders for file types and a folder that's categorically named that contains sdfs.

The calling syntax is different. And if we have a URDF of a robot, why would we need an SDF and vice-versa? This is why I felt having a robot-centric naming convention was not helpful since the directories would just have the one file and unnecessarily add more typing to load the file aka why should a user type `cablerobot/cablerobot.urdf` when they can simply do `kUrdfPath/cablerobot.urdf`?

> Also just realized, why are most of these files except `cerberus...` in a `subt/subT` subfolder?

Yeah I don't know what I was thinking (or not thinking). Updated this in the last commit, thanks!

PR #298, opened 2021-11-02
CI fixes

varunagrawal:
Fixes to the CI

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -11,7 +11,7 @@ jobs:
       CTEST_OUTPUT_ON_FAILURE: ON
       CTEST_PARALLEL_LEVEL: 2
       CMAKE_BUILD_TYPE: ${{ matrix.build_type }}
-      GTSAM_BUILD_UNSTABLE: ${{ matrix.build_unstable }}
+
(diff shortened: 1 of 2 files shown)


varunagrawal:
Turns out the issue was not clang-9 but a recent TBB update we merged in GTSAM. https://github.com/borglab/gtsam/pull/901

danbarla (approved):
GREAT :)

acxz:
> Turns out the issue was not clang-9 but a recent TBB update we merged in GTSAM.

Hello Varun not sure where to comment this, but it seems like this is the best place short of making a new issue. I tried compiling GTDynamics with GTSAM with TBB and without TBB and both compiled fine on my machine. The runtime errors were also the same for me. If there is anything specific related to the gtsam tbb commit or if you have an error log, I can try to look into if you make an issue in this repo and just ping me on it.

varunagrawal:
We undid the TBB change so you may want to apply it again and try. We get intermittent test failures so if you can track down the issue that would be awesome!

(I think it's a race condition but haven't had the time to look into it deeply)

acxz:
Oh yeah forgot it wasn't in the latest `4.1.1` release.

Issue #166, opened 2021-04-21
Different Behavior on Different Machines


yetongumich:
The `testPneumaticFactors.h` do not pass on Varun's machine (before changing the threshold). We should spend some time together to figure out what's the difference in our environment settings that cause such issues.

varunagrawal:
`testPneumaticFactors.h` does not pass in the Linux CI as well. Here's the error log:

```sh
5/47 Test  #5: testPneumaticFactors ....................***Failed    0.03 sec
Not equal:
expected:

  A[p1] = [
	-0.0014666
]
  A[p2] = [
	0.00032725
]
  A[m1] = [
	679972
]
  b = [ 0.623547 ]
  No noise model
actual:

  A[p1] = [
	-0.0014666
]
  A[p2] = [
	0.00032725
]
  A[m1] = [
	677694
]
  b = [ 0.623547 ]
  No noise model
not equal:
expected = [
	679972
]
actual = [
	677694
]
actual - expected = [
	-2277.71
]
Mismatch in Jacobian 3 (base 1), as shown above
/home/varun/borglab/GTDynamics/gtdynamics/jumpingrobot/tests/testPneumaticFactors.cpp:119: Failure: "gtsam::internal::testFactorJacobians(name_, factor, values, diffDelta, 1e-5)" 
There were 1 failures

```

PR #299, opened 2021-11-03
Eigen fixes

varunagrawal:
Since no one seems to be able to figure out the issues wrt different machines (a la #166) I discovered an egregious issue in Joint.h and Link.h, and hopefully that helps.

More details:
1. http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html
2. http://eigen.tuxfamily.org/dox-devel/group__TopicStructHavingEigenMembers.html

### DIFF
--- a/examples/example_spider_walking/main.cpp
+++ b/examples/example_spider_walking/main.cpp
@@ -12,7 +12,6 @@
  * Gandhi, Varun Agrawal
  */
 
-#include <CppUnitLite/TestHarness.h>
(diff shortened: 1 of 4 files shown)


dellaert (approved):
You did not say it in your PR that this actually **solves** an issue, but reading between the lines I guess it does?? Approved on the condition that it is actually needed :-)

varunagrawal:
It does not solve a visible issue, but when I ran valgrind on it to debug something else, the error pertaining to this PR popped up.

PR #301, opened 2021-11-05
Feature/panda

dellaert:
Added Panda URDF. Thanks for your help @gchenfc


models/urdfs/panda/panda.urdf
```diff
@@ -0,0 +1,321 @@
+<?xml version='1.0' encoding='utf-8'?>
+<!-- Manually converted from franka_description/robots/panda_gazebo.xacro -->
```
gchenfc, line 2: Here's the permalink, just for reference/posterity:

https://github.com/frankaemika/franka_ros/blob/8dc80d19aeb2aa88625b1ad42e03d8e8169e3905/franka_description/robots/panda_gazebo.xacro

(correct me if this is _not_ the file this was converted from)

python/tests/test_forward_kinematics.py
```diff
@@ -22,11 +23,12 @@
 
 class TestRobot(GtsamTestCase):
     """Tests related to specific robot config."""
+
     def setUp(self):
         """Set up the fixtures."""
         # load example robot
-        self.ROBOT_MODEL = osp.join(gtd.URDF_PATH, "a1.urdf")
-        self.robot = gtd.CreateRobotFromFile(self.ROBOT_MODEL)
+        model_file = Path(gtd.URDF_PATH) / "a1.urdf"
```
gchenfc, line 30: cool!

python/tests/test_panda.py
```diff
@@ -0,0 +1,53 @@
...
+            [1, 0, 0],
+            [0, -1, 0],
+            [0, 0, -1]
+        ])
+        expected_sT7 = Pose3(sR7, Point3(0.129315, 0.00227401, 0.88387))
+        actual_sT7 = gtd.Pose(fk, 7)
+        self.gtsamAssertEquals(actual_sT7, expected_sT7, tol=1e-3)
+
+
+unittest.main(argv=[''], verbosity=1, exit=False)
```
gchenfc, line 53: not that it makes much of a difference, but could this be placed in `if __name__ == '__main__':` block just in case someone foolishly tries to import this file?

gchenfc (approved):
LGTM!

python/tests/test_panda.py
dellaert: Done

models/urdfs/panda/panda.urdf
dellaert: Correct, but I think this comment is better than a permalink

python/tests/test_forward_kinematics.py
dellaert: Yah!

PR #304, opened 2021-11-10
feature/expression_factors

yetongumich:
Create constrained expression for
- TwistFactor
- TorqueFactor
- WrenchEquivalenceFactor
Notice that for WrenchFactor and TwistAccelerationFactor, there are too many variables involved in the function, and the Expression constructor currently cannot deal with that (it currently only handles unary, binary and tenary functions). I will make a future PR to handle those factors.


gtdynamics/universal_robot/Joint.cpp
```diff
@@ -326,4 +345,50 @@ gtsam::Expression<gtsam::Vector6> Joint::poseConstraint(
...
+
+/* ************************************************************************* */
+gtsam::Vector6_ Joint::wrenchEquivalenceConstraint(uint64_t t) const {
+  gtsam::Vector6_ wrench_p(internal::WrenchKey(parent()->id(), id(), t));
+  gtsam::Vector6_ wrench_c(internal::WrenchKey(child()->id(), id(), t));
+  gtsam::Double_ q(internal::JointAngleKey(id(), t));
+
+  gtsam::Vector6_ wrench_c_hat(
+      std::bind(&Joint::transformWrenchCoordinate, this, child(), std::placeholders::_1,
+                std::placeholders::_2, std::placeholders::_3,
```
dellaert, line 373: Think about simplifying with a function that takes expressions, a la "logmap" we added.
At the very least, say "using std::placeholders" above.

dellaert (approved):
Cooool!

gtdynamics/universal_robot/Joint.cpp
yetongumich: There's no logmap associated with this function, as wrenches are Vector6s quantities, so just subtraction shall work.

I tried with using std::placeholders before, it compiles on mac, but does not work on ubuntu, as it reports that boost also has placeholders, and it got confused about which one to use.

PR #305, opened 2021-11-12
Feature/screw access

dellaert:
Better screw axis access :-)


gtdynamics/universal_robot/Joint.h
```diff
@@ -160,14 +160,23 @@ class Joint : public boost::enable_shared_from_this<Joint> {
...
   const Pose3 &jMp() const { return jMp_; }
 
-  /// Transform from the joint frame to the child's center of mass.
+  /// Return (unchanging) pose of the child link's COM in the joint frame.
   const Pose3 &jMc() const { return jMc_; }
 
-  /// Return screw axis expressed in the specified link frame
-  const Vector6 screwAxis(const LinkSharedPtr &link) const {
+  /// Return pose of child in parent link, at rest.
+  Pose3 pMc() const { return jMp_.between(jMc_); }
```
gchenfc, line 170: I wonder if it would be worth "caching" this as a private variable since it's likely to be called very often

gchenfc (approved):
LGTM

oops I think these were my mistakes, thanks for fixing!

gtdynamics/universal_robot/Joint.h
dellaert: Yeah. Different PR though, and profiling-driven.

PR #308, opened 2021-11-16
wrench factor with expressions

yetongumich:
Re-writing wrench factor and related factors using expressions, so we can easily transition to constrained optimization later. This is transitional, because eventually all "soft" factors will disappear and instead be generated by a call to `createFactor` of the `EqualityConstraint` class.

Added Python tests to factors.


dellaert (changes requested):
Kill all comments, finalize, and then re-request review.

gtdynamics/statics/StaticsSlice.cpp
```diff
@@ -53,8 +53,8 @@ gtsam::NonlinearFactorGraph Statics::wrenchPlanarFactors(
   gtsam::NonlinearFactorGraph graph;
   if (p_->planar_axis)
     for (auto&& joint : robot.joints()) {
-      graph.emplace_shared<WrenchPlanarFactor>(
-          p_->planar_cost_model, *p_->planar_axis, joint, slice.k);
+      graph.add(WrenchPlanarFactor(p_->planar_cost_model, *p_->planar_axis,
```
dellaert, line 56: This is a copy! Make sure createFactor uses `boost::create_shared`

gtdynamics/universal_robot/Joint.cpp
```diff
@@ -357,6 +372,32 @@ gtsam::Vector6_ Joint::twistConstraint(uint64_t t) const {
   return twist_c_hat - twist_c;
 }
 
+/* ************************************************************************* */
+gtsam::Vector6_ Joint::twistAccelConstraint(uint64_t t) const {
+  // gtsam::Vector6_ twist_p(internal::TwistKey(parent()->id(), t));
```
dellaert, line 377: ?

gtdynamics/universal_robot/Joint.cpp
```diff
@@ -357,6 +372,32 @@ gtsam::Vector6_ Joint::twistConstraint(uint64_t t) const {
   return twist_c_hat - twist_c;
 }
 
+/* ************************************************************************* */
+gtsam::Vector6_ Joint::twistAccelConstraint(uint64_t t) const {
```
dellaert, line 376: Might be OK if g(x) is the "constraint function", and we can vaguely abuse notation here and call it "Constraint", otherwise too long.

gtdynamics/universal_robot/Joint.cpp
```diff
@@ -140,6 +133,28 @@ Vector6 Joint::transformTwistAccelTo(
...
+    gtsam::OptionalJacobian<6, 1> H_q_ddot,
+    gtsam::OptionalJacobian<6, 6> H_this_twist) const {
+  Vector6 this_twist_ = this_twist ? *this_twist : Vector6::Zero();
+  Vector6 screw_axis_ = isChildLink(link) ? cScrewAxis_ : pScrewAxis_;
+
+  // i = other link
+  // j = this link
+  auto other = otherLink(link);
+
+  Vector6 this_twist_accel =
```
dellaert, line 154: where is this used?

gtdynamics/universal_robot/Joint.cpp
```diff
@@ -357,6 +372,32 @@ gtsam::Vector6_ Joint::twistConstraint(uint64_t t) const {
...
+gtsam::Vector6_ Joint::twistAccelConstraint(uint64_t t) const {
+  // gtsam::Vector6_ twist_p(internal::TwistKey(parent()->id(), t));
+  gtsam::Vector6_ twist_c(internal::TwistKey(child()->id(), t));
+  gtsam::Vector6_ twistAccel_p(internal::TwistAccelKey(parent()->id(), t));
+  gtsam::Vector6_ twistAccel_c(internal::TwistAccelKey(child()->id(), t));
+  gtsam::Double_ q(internal::JointAngleKey(id(), t));
+  gtsam::Double_ qVel(internal::JointVelKey(id(), t));
+  gtsam::Double_ qAccel(internal::JointAccelKey(id(), t));
+
+  gtsam::Vector6_ twistAccel_c_hat1(
```
dellaert, line 385: Consider lambdas with exactly right arguments, no placeholder needs.

gtdynamics/universal_robot/Joint.h
```diff
@@ -286,15 +286,23 @@ class Joint : public boost::enable_shared_from_this<Joint> {
    * accel and a Values object containing this joint's angle Value and
    * derivatives.
    */
-  gtsam::Vector6 transformTwistAccelTo(
-      const LinkSharedPtr &link, double q, double q_dot, double q_ddot,
-      boost::optional<gtsam::Vector6> this_twist = boost::none,
+  gtsam::Vector6 transformTwistAccelTo1(
```
dellaert, line 289: Maybe have them disappear if not used, as lambdas.

gtdynamics/universal_robot/Link.cpp
```diff
@@ -0,0 +1,75 @@
...
+gtsam::Vector6_ Link::wrenchConstraint(
+    const std::vector<DynamicsSymbol>& wrench_keys, uint64_t t,
+    const boost::optional<gtsam::Vector3>& gravity) const {
+  // Collect wrenches to implement L&P Equation 8.48 (F = ma)
+  std::vector<gtsam::Vector6_> wrenches;
+
+  // Coriolis forces.
+  gtsam::Vector6_ twist(internal::TwistKey(id(), t));
+  gtsam::Vector6_ wrench_coriolis(
+      std::bind(Coriolis, inertiaMatrix(), std::placeholders::_1,
```
dellaert, line 37: elegance detector firing

gtdynamics/utils/JsonSaver.h
```diff
@@ -225,21 +225,7 @@ class JsonSaver {
...
-    } else if (dynamic_cast<const TwistAccelFactor*>(&(*factor))) {
-      return "TwistAccel";
-    // } else if (dynamic_cast<const TorqueFactor*>(&(*factor))) {
-    //   return "Torque";
-    } else if (dynamic_cast<const WrenchPlanarFactor*>(&(*factor))) {
-      return "WrenchPlanar";
-    // } else if (dynamic_cast<const WrenchEquivalenceFactor*>(&(*factor))) {
-    //   return "WrenchEq";
-    } else if (dynamic_cast<const gtsam::PriorFactor<double>*>(&(*factor))) {
+    if (dynamic_cast<const gtsam::PriorFactor<double>*>(&(*factor))) {
```
dellaert, line 228: Add a TODO to use RTTI to do this generically.

gtdynamics.i
```diff
@@ -86,16 +86,9 @@ gtsam::NonlinearFactor TwistFactor(const gtsam::noiseModel::Base *cost_model,
...
-  TwistAccelFactor(gtsam::Key twist_key_c, gtsam::Key twistAccel_key_p,
-                   gtsam::Key twistAccel_key_c, gtsam::Key q_key,
-                   gtsam::Key qVel_key, gtsam::Key qAccel_key,
-                   const gtsam::noiseModel::Base *cost_model,
-                   const gtdynamics::Joint *joint);
-
-  void print(const string &s="",
-             const gtsam::KeyFormatter &keyFormatter=gtdynamics::GTDKeyFormatter);
-};
+gtsam::NonlinearFactor TwistAccelFactor(const gtsam::noiseModel::Base *cost_model,
```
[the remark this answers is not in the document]
varunagrawal: @dellaert this won't work. `TwistAccelFactor` returns an `ExpressionFactor` which is a type that is not wrapped. The moment you invoke the function from Python (or any wrapped function), the interpreter will crap out with an unknown type error.

Getting ExpressionFactor wrapped is very hard because it requires us to wrap Expression in general and that requires a lot of work with functionals. I don't know if you already looked into the feasibility of this from a wrapping perspective when you decided to make these changes.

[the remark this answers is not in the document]
yetongumich: That makes sense. I'll try to make it return a boost::shared_ptr instead

gtdynamics/dynamics/Dynamics.h
```diff
@@ -24,4 +24,14 @@ gtsam::Vector6 Coriolis(const gtsam::Matrix6 &inertia,
                         const gtsam::Vector6 &twist,
                         gtsam::OptionalJacobian<6, 6> H_twist = boost::none);
 
+/// Matrix vector multiplication.
+gtsam::Vector6 MatVecMult(const gtsam::Matrix6 &inertia,
```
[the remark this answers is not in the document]
yetongumich: Done

gtdynamics/dynamics/Dynamics.cpp
```diff
@@ -40,4 +40,22 @@ Vector6 Coriolis(const Matrix6 &inertia, const Vector6 &twist,
   return result;
 }
 
+Vector6 MatVecMult(const Matrix6 &inertia, const Vector6 &twist,
```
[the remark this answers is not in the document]
yetongumich: Done

gtdynamics/universal_robot/Joint.h
yetongumich: Done

gtdynamics/universal_robot/Joint.cpp
yetongumich: done

gtdynamics/universal_robot/Joint.cpp
yetongumich: done. changed to lambda functions

gtdynamics/utils/JsonSaver.h
yetongumich: Done

gtdynamics/universal_robot/Link.cpp
yetongumich: I don't get it. Is the function too long?

gtdynamics/factors/WrenchFactor.h
```diff
@@ -33,61 +36,34 @@ namespace gtdynamics {
...
+/**
+ * Function to use for Python wrapper to add WrenchFactor, since
+ * ExpressionFactors cannot currently be used in Python.
+ */
+inline void addWrenchFactor(
+    gtsam::NonlinearFactorGraph &graph,
+    const gtsam::SharedNoiseModel &cost_model, const LinkConstSharedPtr &link,
+    const std::vector<DynamicsSymbol> &wrench_keys, int time,
+    const boost::optional<gtsam::Vector3> &gravity = boost::none) {
+  graph.add(gtsam::ExpressionFactor<gtsam::Vector6>(
```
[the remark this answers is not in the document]
yetongumich: Done

gtdynamics/universal_robot/Joint.cpp
yetongumich: Done

gtdynamics/statics/StaticsSlice.cpp
yetongumich: Done. Make WrenchPlanarFactor and all other factor-creation functions return a shared_ptr to avoid copy.

gtdynamics/universal_robot/Link.cpp
dellaert: Same as with the other expressions. If Coriolis was a functor, you wouldn’t have to do this placeholder dance

gtdynamics/factors/WrenchPlanarFactor.h
```diff
@@ -28,78 +31,39 @@
...
   }
 
-  /// @return a deep copy of this factor
-  gtsam::NonlinearFactor::shared_ptr clone() const override {
-    return boost::static_pointer_cast<gtsam::NonlinearFactor>(
-        gtsam::NonlinearFactor::shared_ptr(new This(*this)));
-  }
+  auto wrench_key = internal::WrenchKey(joint->child()->id(), joint->id(), k);
+  gtsam::Vector6_ wrench(wrench_key);
+  gtsam::Vector3_ error(std::bind(MatVecMult36, H_wrench, std::placeholders::_1,
```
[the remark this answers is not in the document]
yetongumich: Done

gtdynamics/universal_robot/Joint.cpp
yetongumich: Let's call it "Constraint" as it is now.

gtdynamics.i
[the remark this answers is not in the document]
yetongumich: make the functions return NonlinearFactor::shared_ptr. In this way, we won't expose the ExpressionFactor to the Python wrapper, and the wrapper works just fine.

gtdynamics/cablerobot/src/cdpr_planar.py
```diff
@@ -151,18 +151,14 @@ def dynamics_factors(self, ks=[]):
         """
         dfg = gtsam.NonlinearFactorGraph()
         for k in ks:
-            wf = gtd.WrenchFactor(
-                    gtd.internal.TwistKey(self.ee_id(), k).key(),
-                    gtd.internal.TwistAccelKey(self.ee_id(), k).key(),
+            gtd.addWrenchFactor(dfg,
```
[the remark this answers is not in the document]
yetongumich: Done

gtdynamics.i
```diff
@@ -112,28 +105,26 @@ class MinTorqueFactor : gtsam::NonlinearFactor {
...
-                const boost::optional<gtsam::Vector3> &gravity);
-  void print(const string &s="",
-             const gtsam::KeyFormatter &keyFormatter=gtdynamics::GTDKeyFormatter);
-};
+gtsam::NonlinearFactor* WrenchFactor(
+    const gtsam::noiseModel::Base *cost_model, const gtdynamics::Link *link,
+    const std::vector<gtdynamics::DynamicsSymbol> wrench_keys, int t = 0,
+    const boost::optional<gtsam::Vector3> &gravity);
+
+void addWrenchFactor(
```
dellaert, line 113: so, do we still need this then?

gtdynamics/factors/WrenchFactor.h
```diff
@@ -33,61 +36,33 @@ namespace gtdynamics {
...
-  void serialize(ARCHIVE &ar, const unsigned int version) {  // NOLINT
-    ar &boost::serialization::make_nvp(
-        "NoiseModelFactor", boost::serialization::base_object<Base>(*this));
-  }
-};
+/**
+ * Function to use for Python wrapper to add WrenchFactor, since
+ * ExpressionFactors cannot currently be used in Python.
+ */
+inline void addWrenchFactor(
```
dellaert, line 59: still needed??

gtdynamics/universal_robot/Joint.cpp
```diff
@@ -357,6 +314,60 @@ gtsam::Vector6_ Joint::twistConstraint(uint64_t t) const {
...
+             gtsam::OptionalJacobian<6, 1> H_q,
+             gtsam::OptionalJacobian<6, 6> H_other_twist_accel) {
+        Pose3 jTi = relativePoseOf(parent(), q);
+        Vector6 this_twist_accel = jTi.AdjointMap() * other_twist_accel;
+
+        if (H_other_twist_accel) {
+          *H_other_twist_accel = jTi.AdjointMap();
+        }
+        if (H_q) {
+          // TODO(frank): really, zero below? Check derivatives. Also,
```
dellaert, line 340: Create issue to check this (or solve it by checking derivatives)

gtdynamics/universal_robot/Joint.cpp
```diff
@@ -357,6 +314,60 @@ gtsam::Vector6_ Joint::twistConstraint(uint64_t t) const {
...
+            cScrewAxis_ * q_ddot;
+        if (H_q_dot) {
+          *H_q_dot = Pose3::adjointMap(this_twist) * cScrewAxis_;
+        }
+        if (H_q_ddot) {
+          *H_q_ddot = cScrewAxis_;
+        }
+        return this_twist_accel;
+      };
+  gtsam::Vector6_ twistAccel_c_hat2(transformTwistAccelTo2, qVel, qAccel, twist_c);
```
dellaert, line 366: nice !

gtdynamics/universal_robot/Link.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  // Calculate resultant wrench.
+  gtsam::Vector6_ error(gtsam::Z_6x1);
+  for (const auto& wrench : wrenches) {
+    error += wrench;
+  }
+  return error;
+}
+
+}  // namespace gtdynamics
```
dellaert, line 75: missing newline at end of file

dellaert (approved):
Nice. Still a few nits...

gtdynamics.i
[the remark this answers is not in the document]
varunagrawal: That won't work either because a NonlinearFactor is an abstract base type. You need to return the actual type of the return value.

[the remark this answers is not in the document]
varunagrawal: @yetongumich I recommend adding a Python unit test rather than shooting arrows in the dark. This will ensure your changes are ironclad.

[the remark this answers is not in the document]
dellaert: I actually think this will work. As long it is a nonlinear factor shared pointer. But I don’t understand why we need them. I’d rather remove these factors from the wrapper. They are on the way out anyway.

@varunagrawal , you can use the temporary factor in your other PR.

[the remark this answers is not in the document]
dellaert: But in general I agree with @varunagrawal  comment, @yetongumich : have unit tests.

gtdynamics.i
```diff
@@ -81,24 +81,17 @@ class ContactPointFactor : gtsam::NoiseModelFactor {
 };
 
 #include <gtdynamics/factors/TwistFactor.h>
-gtsam::NonlinearFactor TwistFactor(const gtsam::noiseModel::Base *cost_model,
+gtsam::NonlinearFactor* TwistFactor(const gtsam::noiseModel::Base *cost_model,
```
varunagrawal, line 84: We need a Python unit test verifying this works before this can be merged.

varunagrawal (changes requested):

gtdynamics.i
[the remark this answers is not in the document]
varunagrawal: Yetong mentioned using a NonlinearFactor but he actually is returning a NoiseModelFactor. That works because it is not an abstract base type, so his comment is misleading aka not cool bro.

@dellaert what do you mean by these factors are on their way out? Can you please elaborate on this since this was not mentioned to the rest of the group. How would this impact being able to construct specific factor graphs for various problems (and not just state estimation)?

gtdynamics/factors/PoseFactor.h
```diff
@@ -46,9 +43,10 @@ using boost::assign::cref_list_of;
  * @param joint The joint connecting the two poses.
  * @param time The timestep at which this factor is defined.
  */
-inline gtsam::ExpressionFactor<gtsam::Vector6> PoseFactor(const gtsam::SharedNoiseModel &cost_model,
-             const JointConstSharedPtr &joint, int time) {
-  return gtsam::ExpressionFactor<gtsam::Vector6>(
+inline gtsam::ExpressionFactor<gtsam::Vector6>::shared_ptr PoseFactor(
```
varunagrawal, line 46: This should also be updated.

gtdynamics/factors/PoseFactor.h
```diff
@@ -63,13 +61,13 @@ inline gtsam::ExpressionFactor<gtsam::Vector6> PoseFactor(const gtsam::SharedNoi
...
    * @param joint The joint connecting the two poses
    */
-inline gtsam::ExpressionFactor<gtsam::Vector6>  PoseFactor(DynamicsSymbol wTp_key, DynamicsSymbol wTc_key,
-             DynamicsSymbol q_key,
-             const gtsam::noiseModel::Base::shared_ptr &cost_model,
-             JointConstSharedPtr joint) {
-  return gtsam::ExpressionFactor<gtsam::Vector6>(
-             cost_model, gtsam::Vector6::Zero(),
-             joint->poseConstraint(wTp_key.time()));
+inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
```
varunagrawal, line 64: Is it the intention to return this as a NoiseModelFactor? Now the type info as an ExpressionFactor is lost so this may have repercussions that I am not sure of. @dellaert comments?

gtdynamics/factors/TorqueFactor.h
```diff
@@ -42,11 +42,11 @@ namespace gtdynamics {
...
  * @param joint JointConstSharedPtr to the joint
  */
-inline gtsam::ExpressionFactor<double> TorqueFactor(
+inline gtsam::NoiseModelFactor::shared_ptr TorqueFactor(
     const gtsam::noiseModel::Base::shared_ptr &cost_model,
     const JointConstSharedPtr &joint, size_t k = 0) {
-  return gtsam::ExpressionFactor<double>(cost_model, 0.0,
-                                         joint->torqueConstraint(k));
+  return boost::make_shared<gtsam::ExpressionFactor<double>>(
+      cost_model, 0.0, joint->torqueConstraint(k));
```
varunagrawal, line 49: Hardcoding? That's usually a big no. Any reason this isn't just an optional function parameter?

varunagrawal (changes requested):

gtdynamics/factors/PoseFactor.h
varunagrawal: One (debatable) downside I've noticed is that the test code now needs to use pointer syntax. So `factor.evaluateError` doesn't work and needs to be `factor->evaluateError`. Personally I am not a big fan because everywhere else, we use the former syntax and mixing-and-matching never goes well (aka greater cognitive overload aka less user friendly).

dellaert: These factors are on the way out. Let's stop spending time on them.

gtdynamics.i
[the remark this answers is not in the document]
yetongumich: @varunagrawal Yes. I was initially returning NonlinearFactor::shared_ptr, but I later discovered that some unit tests require some methods in NoiseModelFactor, so I returned NoiseModelFactor::shared_ptr instead. They python wrapper works, but I'll add the unit tests as well.

gtdynamics.i
yetongumich: Removed

gtdynamics/factors/WrenchFactor.h
yetongumich: removed

gtdynamics/universal_robot/Link.cpp
yetongumich: Done

gtdynamics/factors/TorqueFactor.h
yetongumich: I don't get it. Is it the 0.0 that you are referring to? The torqueConstraint returns an expression, and the constraint itself basically forces the expression to be 0.

gtdynamics/universal_robot/Joint.cpp
yetongumich: The derivatives are checked in testTwistAccelerationFactor. So it should be fine.

gtdynamics.i
yetongumich: Python tests added

gtdynamics/factors/PoseFactor.h
varunagrawal: And what will we be replacing them with? Or what is the expected use case? You can't simply make a blanket statement like that without considering the dependencies of others. :slightly_smiling_face:

gtdynamics.i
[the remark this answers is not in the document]
varunagrawal: What do you mean by the python wrapper works? That it compiles successfully? That is not the same as it working.

gtdynamics/factors/TorqueFactor.h
varunagrawal: Gotcha. Then that should be a comment `The expression constrains the forces to be 0`.

varunagrawal:
@yetongumich CI is failing due to both C++ and Python.

python/tests/test_factors.py
```diff
@@ -34,3 +34,63 @@ def test_revolute_joint_measurement_factor(self):
...
+    
+        self.values = gtsam.Values()
+        gtd.InsertJointAngle(self.values, 0, self.k, 0.0)
+        gtd.InsertJointAngle(self.values, 1, self.k, 0.0)
+        self.values = self.robot.forwardKinematics(self.values, self.k, "link_0")
+        gtd.InsertTorque(self.values, 0, self.k, 0.0)
+        gtd.InsertWrench(self.values, 0, 0, self.k, np.zeros(6))
+        gtd.InsertWrench(self.values, 1, 0, self.k, np.zeros(6))
+
+    def test_pose_factor(self):
```
varunagrawal, line 58: These should ideally be their own class for each factor so we can have methods for construction, error evaluation and jacobian computation.

varunagrawal: I already started working on that, so you can use this as a template. Note that I want the constructor tests in particular since that exposes a lot of bugginess in the wrapping process.

```python
"""
 * GTDynamics Copyright 2020, Georgia Tech Research Corporation,
 * Atlanta, Georgia 30332-0415
 * All Rights Reserved
 * See LICENSE for the license information
 *
 * @file  test_dynamics_graph.py
 * @brief Unit tests for dynamics graph.
 * @author Varun Agrawal
"""

import os.path as osp
import unittest

import gtdynamics as gtd
import gtsam
import numpy as np


class TestFactors(unittest.TestCase):
    """
    Base class for testing various factors.
    Provides needed fixtures and common functions.
    """
    def setUp(self):
        self.k = 0
        self.wTp_key = gtd.internal.PoseKey(0, self.k).key()
        self.wTc_key = gtd.internal.PoseKey(1, self.k).key()

        ROBOT_FILE = osp.join(gtd.SDF_PATH, "test", "simple_rr.sdf")
        self.robot = gtd.CreateRobotFromFile(str(ROBOT_FILE), "simple_rr_sdf")


class TestJointMeasurementFactor(TestFactors):
    """Test suite for various versions of the JointMeasurementFactor."""
    def test_revolute_joint_measurement_factor(self):
        """Test RevoluteJointMeasurementFactor."""
        factor = gtd.JointMeasurementFactor(
            gtsam.noiseModel.Isotropic.Sigma(6, 0.1),
            self.robot.joint("joint_1"), np.pi / 4, self.k)

        self.assertIsInstance(factor, gtd.JointMeasurementFactor)


class TestTempPoseFactor(TestFactors):
    """Test suite for various versions of the JointMeasurementFactor."""
    def test_temp_pose_factor_constructor(self):
        """Test constructor."""
        joint = self.robot.joint("joint_1")
        joint_key = gtd.internal.JointAngleKey(joint.id(), self.k).key()
        pose_model = gtsam.noiseModel.Isotropic.Sigma(6, 0.1)
        pose_factor = gtd.TempPoseFactor(self.wTp_key, self.wTc_key, joint_key,
                                         pose_model, joint)

        self.assertIsInstance(pose_factor, gtd.TempPoseFactor)

    def test_pose_factor_constructor(self):
        """Test constructor."""
        joint = self.robot.joint("joint_1")
        joint_key = gtd.internal.JointAngleKey(joint.id(), self.k).key()
        pose_model = gtsam.noiseModel.Isotropic.Sigma(6, 0.1)
        pose_factor = gtd.PoseFactor(self.wTp_key, self.wTc_key, joint_key,
                                     pose_model, joint)

        self.assertIsInstance(pose_factor, gtd.PoseFactor)


class TestContactKinematicsAccelFactor(TestFactors):
    def test_constructor(self):
        pass


class TestMinTorqueFactor(TestFactors):
    def test_constructor(self):
        pass


class TestPreintegratedContactFactors(TestFactors):
    def test_constructor(self):
        pass


class TestWrenchFactor(TestFactors):
    def test_constructor(self):
        pass


class TestCollocationFactors(TestFactors):
    def test_constructor(self):
        pass


class TestContactKinematicsTwistFactor(TestFactors):
    def test_constructor(self):
        pass


class TestTorqueFactor(TestFactors):
    def test_constructor(self):
        pass


class TestWrenchPlanarFactor(TestFactors):
    def test_constructor(self):
        pass


class TestContactDynamicsFrictionConeFactor(TestFactors):
    def test_constructor(self):
        pass


class TestContactPointFactor(TestFactors):
    def test_constructor(self):
        pass


class TestObjectiveFactors(TestFactors):
    def test_constructor(self):
        pass


class TestTwistAccelFactor(TestFactors):
    def test_constructor(self):
        pass


class TestContactDynamicsMomentFactor(TestFactors):
    def test_constructor(self):
        pass


class TestForwardKinematicsFactor(TestFactors):
    def test_constructor(self):
        pass


class TestTwistFactor(TestFactors):
    def test_constructor(self):
        pass


class TestContactEqualityFactor(TestFactors):
    def test_constructor(self):
        pass


class TestJointLimitFactor(TestFactors):
    def test_constructor(self):
        pass


class TestPointGoalFactor(TestFactors):
    def test_constructor(self):
        pass


class TestWrenchEquivalenceFactor(TestFactors):
    def test_constructor(self):
        pass


class TestContactHeightFactor(TestFactors):
    def test_constructor(self):
        pass

```

varunagrawal (changes requested):

gtdynamics/factors/PoseFactor.h
dellaert: Varun, we have talked about this several times now: equalityConstints.If you’re not getting it yet, we can talk about it next Monday. For now, I want *all* those  factors gone from the wrapper unless there is code using them. AFAIK tell this is *only* you, formPoseFactor. correct me if I am wrong.

gtdynamics.i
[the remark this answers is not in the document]
dellaert: @varunagrawal please meet me on Bluejeans now if you can.
@yetongumich please don't add any unit tests for now until I understand Varun's objection, after he has the whole context.

dellaert:
All this extra work for nothing! @varunagrawal I am rather upset about this.

varunagrawal:
@yetongumich Frank and I spoke about this PR and the fix is easy. We just need to remove the wrapped functions for all the factors (since that is the root cause of the breakage) and then you can also kill the python unit tests for the factors. Once you do that, this PR will be awesome and I will gladly approve. :)

dellaert:
Thanks @varunagrawal . For all the "factor functions". These functions were stopgap measures to make the c++ code work (almost) as is, and the corresponding factors were not used (except PoseFactor, other PR). Sorry about the extra work, @yetongumich , I tried to say this in our meeting and in the PR comments but I guess I there was a breakdown in communications.

varunagrawal:
One more thing before I forget. We can undo the change from ExpressionFactor to NoiseModelFactor::shared_ptr that was done due to my comments. Since the wrapper is not an issue anymore, having full type info of the ExpressionFactor will be fantastic!

gtdynamics/cablerobot/src/cdpr_planar.py
```diff
@@ -151,18 +151,14 @@ def dynamics_factors(self, ks=[]):
...
                     [
                         gtd.internal.WrenchKey(self.ee_id(), 0, k),
                         gtd.internal.WrenchKey(self.ee_id(), 1, k),
                         gtd.internal.WrenchKey(self.ee_id(), 2, k),
                         gtd.internal.WrenchKey(self.ee_id(), 3, k)
-                    ],
-                    gtd.internal.PoseKey(self.ee_id(), k).key(),
-                    self.costmodel_wrench, self.eelink().inertiaMatrix(), self.params.gravity)
-            dfg.push_back(wf)
+                    ], k, self.params.gravity))
```
gchenfc, line 161: Thanks for the ping.  Just for context, the primary reason I use these factors is because my CableFactors (`CableLengthFactor`, `CableVelocityFactor`, `CableTensionFactor`) are not quite the same as `PoseFactor`/`TwistFactor`/`TwistAccelFactor`/`TorqueFactor` so therefore I cannot create a "Cable Joint" and therefore cannot use the `Robot` machinery :( figuring out a nice way to incorporate these is on my TODO list.

After your guys' long discussion, it seems the conclusion is that WrenchFactor is going to no longer be part of the python interface?  If so, could you maybe create a `TempWrenchFactor` similar to `TempPoseFactor` to tide me over until equality stuff is ready for me to migrate everything over?

Also, I _think_ it could also be something simple and wouldn't need to retain the entire old factor as in `TempPoseFactor`.  Something like:
```c++
gtsam::NonlinearFactor::shared_ptr TempWrenchFactor(args) {
    return boost::make_shared<WrenchFactor>(args);
}
```
gtdynamics.i:
```c++
gtsam::NonlinearFactor::shared_ptr TempWrenchFactor(args);
```
or something (not tested)

dellaert: I suggest meeting briefly meeting with Yetong

yetongumich:
- Got all factor wrappers removed (except for a temporary wrapper for WrenchFactor which keeps the cablerobot tests still working).
- Added a TODO and issue in removing the wrapper for WrenchFactor in a later PR. 
- As for whether the factor functions should return an Expression factor or a shared_ptr, Frank and I had this discussion before, and it's still reasonable to return a shared_ptr, because this will avoid the copy of the factor when adding to the factor graph.

gtdynamics/cablerobot/src/cdpr_planar.py
yetongumich: A temporary solution is to keep the WrenchFactor wrapper to make the test working, and making an Issue and a TODO to resolve that in the future.

python/tests/test_factors.py
yetongumich: temporarily remove all the factor tests and wrapper, except for the WrenchFactor to make cablerobot tests working

gtdynamics/factors/TorqueFactor.h
yetongumich: It's a misunderstanding. The `torqueConstraint` and all other such functions returns an expression g(x). And the EqualityConstraint to create is basically g(x)=0.

gtdynamics/universal_robot/Link.cpp
yetongumich: Added a TODO here

varunagrawal (approved):

varunagrawal:
@yetongumich we've merged this which is awesome, but we should still revert all the `NoiseModelFactor::shared_ptr`s to `ExpressionFactor`s.

PR #311, opened 2021-11-22
A1 benchmark example

danbarla:
This PR adds an A1 example in order to create a benchmark for future improvements on the factor graph.
The example includes a main.cpp which builds the factor graph, optimizes and writes the results to a .csv file.
The example inclues also a python simulation script which takes the values and simulates the robot motion.


danbarla:
https://user-images.githubusercontent.com/67676161/143093504-359ff642-10b0-4ce0-8237-9d73267fd087.mp4

examples/example_a1_walking/main.cpp
```diff
@@ -0,0 +1,172 @@
...
+using gtsam::Point3;
+using gtsam::Pose3;
+using gtsam::Rot3;
+using gtsam::Vector6;
+using gtsam::noiseModel::Isotropic;
+using gtsam::noiseModel::Unit;
+
+using namespace gtdynamics;
+
+// Returns a Trajectory object for a single spider walk cycle.
```
dellaert, line 40: Spider? ;-)

dellaert (approved):
LGTM, but I thought we already had A1 files. Maybe in SDF models??

examples/example_a1_walking/main.cpp
danbarla: I had a little trouble with the a1 sdf which was in the repo already. It wouldn't load in pybullet at all.

dellaert (approved):
For future reference, you don't need to request a re-review if the changes were simple to implement and I already approved:-)

PR #313, opened 2021-11-23
Move jumping robot python files to jumpingrobot directory

varunagrawal:
The files are not example files, rather source and test files for Python, hence they should be in the correct directory, akin to the `cablerobot`.


varunagrawal:
The jumpingrobot python unit tests are never tested by the CI. Adding them in a support PR.

dellaert:
Up to @yetongumich to review

yetongumich (approved):

PR #316, opened 2021-11-30
Robot Serialization

varunagrawal:
This class provides serialization capabilities for the Robot class and all the currently defined links and joints.

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -17,12 +17,12 @@ jobs:
       matrix:
         # Github Actions requires a single row to be added to the build matrix.
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
-        name: [ubuntu-20.04-gcc-9, ubuntu-20.04-clang-10]
+        name: [ubuntu-18.04-gcc-9, ubuntu-20.04-clang-10]
 
         build_type: [Debug, Release]
         include:
-          - name: ubuntu-20.04-gcc-9
-            os: ubuntu-20.04
+          - name: ubuntu-18.04-gcc-9
+            os: ubuntu-18.04
             compiler: gcc
             version: "9"
 
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -17,17 +17,17 @@ jobs:
       matrix:
         # Github Actions requires a single row to be added to the build matrix.
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
-        name: [ubuntu-18.04-gcc-9, ubuntu-18.04-clang-10]
+        name: [ubuntu-20.04-gcc-9, ubuntu-20.04-clang-10]
 
         build_type: [Debug, Release]
         include:
-          - name: ubuntu-18.04-gcc-9
-            os: ubuntu-18.04
+          - name: ubuntu-20.04-gcc-9
+            os: ubuntu-20.04
             compiler: gcc
             version: "9"
 
-          - name: ubuntu-18.04-clang-10
-            os: ubuntu-18.04
+          - name: ubuntu-20.04-clang-10
+            os: ubuntu-20.04
             compiler: clang
             version: "10"
 
--- a/gtdynamics/universal_robot/Joint.h
+++ b/gtdynamics/universal_robot/Joint.h
@@ -61,6 +61,17 @@ struct JointScalarLimit {
   double value_lower_limit = -M_PI_2;
   double value_upper_limit = M_PI_2;
   double value_limit_threshold = 1e-9;
+
+  JointScalarLimit() {}
+
+  /** Serialization function */
+  friend class boost::serialization::access;
+  template <class ARCHIVE>
+  void serialize(ARCHIVE &ar, const unsigned int /*version*/) {
+    ar &BOOST_SERIALIZATION_NVP(value_lower_limit);
+    ar &BOOST_SERIALIZATION_NVP(value_upper_limit);
+    ar &BOOST_SERIALIZATION_NVP(value_limit_threshold);
+  }
 };
 
 /**
@@ -85,6 +96,8 @@ struct JointParams {
   friend class boost::serialization::access;
   template <class ARCHIVE>
   void serialize(ARCHIVE &ar, const unsigned int /*version*/) {
+    ar &BOOST_SERIALIZATION_NVP(effort_type);
+    ar &BOOST_SERIALIZATION_NVP(scalar_limits);
     ar &BOOST_SERIALIZATION_NVP(velocity_limit);
     ar &BOOST_SERIALIZATION_NVP(velocity_limit_threshold);
     ar &BOOST_SERIALIZATION_NVP(acceleration_limit);
--- a/gtdynamics/universal_robot/Joint.h
+++ b/gtdynamics/universal_robot/Joint.h
@@ -20,9 +20,9 @@
 #include <gtsam/geometry/Pose3.h>
 #include <gtsam/linear/GaussianFactorGraph.h>
 #include <gtsam/nonlinear/Expression.h>
-#include <gtsam/nonlinear/expressions.h>
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
 #include <gtsam/nonlinear/Values.h>
+#include <gtsam/nonlinear/expressions.h>
 
 #include <boost/enable_shared_from_this.hpp>
 #include <boost/shared_ptr.hpp>
@@ -80,6 +80,20 @@ struct JointParams {
 
   /// Constructor
   JointParams() {}
+
+  /** Serialization function */
+  friend class boost::serialization::access;
+  template <class ARCHIVE>
+  void serialize(ARCHIVE &ar, const unsigned int /*version*/) {
+    ar &BOOST_SERIALIZATION_NVP(velocity_limit);
+    ar &BOOST_SERIALIZATION_NVP(velocity_limit_threshold);
+    ar &BOOST_SERIALIZATION_NVP(acceleration_limit);
+    ar &BOOST_SERIALIZATION_NVP(acceleration_limit_threshold);
+    ar &BOOST_SERIALIZATION_NVP(torque_limit);
+    ar &BOOST_SERIALIZATION_NVP(torque_limit_threshold);
+    ar &BOOST_SERIALIZATION_NVP(damping_coefficient);
+    ar &BOOST_SERIALIZATION_NVP(spring_coefficient);
+  }
 };
 
 /// Joint is the base class for a joint connecting two Link objects.
@@ -212,6 +226,10 @@ class Joint : public boost::enable_shared_from_this<Joint> {
   }
 
   bool operator!=(const Joint &other) const { return !(*this == other); }
+
+  bool equals(const Joint &other, double tol = 0) const {
+    return *this == other;
+  }
 
   friend std::ostream &operator<<(std::ostream &stream, const Joint &j);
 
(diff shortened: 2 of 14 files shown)


dellaert (commented):
I trust @ProfFan to review

tests/testRevoluteJoint.cpp
```diff
@@ -23,21 +24,7 @@
...
-using gtsam::numericalDerivative22;
-using gtsam::Point3;
-using gtsam::Pose3;
-using gtsam::Rot3;
-using gtsam::Values;
-using gtsam::Vector;
-using gtsam::Vector3;
-using gtsam::Vector6;
-
+using namespace gtsam;
```
ProfFan, line 27: same

ProfFan (changes requested):

tests/testPrismaticJoint.cpp
```diff
@@ -22,7 +23,7 @@
 #include "gtdynamics/utils/utils.h"
 
 using namespace gtdynamics;
-using gtsam::assert_equal, gtsam::Pose3, gtsam::Point3, gtsam::Rot3;
+using namespace gtsam;
```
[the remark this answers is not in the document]
varunagrawal: These don't quite matter for tests since tests are encapsulated and we can work with including the whole namespace. Do you still want me to add them?

.github/workflows/linux-ci.yml
```diff
@@ -17,17 +17,17 @@ jobs:
       matrix:
         # Github Actions requires a single row to be added to the build matrix.
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
-        name: [ubuntu-18.04-gcc-9, ubuntu-18.04-clang-10]
+        name: [ubuntu-20.04-gcc-9, ubuntu-20.04-clang-10]
```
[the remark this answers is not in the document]
varunagrawal: 18.04 has Boost 1.65 and not the minimum 1.67 that GTSAM supports. As such, that serialization bug that we had about deepcopies rears its head again with 18.04.

We'll have to do a custom install of Boost if you want 18.04 which will significantly increase CI time. Thoughts?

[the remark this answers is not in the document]
dellaert: Help, we lost 18.04 for GTSAM???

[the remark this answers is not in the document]
varunagrawal: Nope. We do a custom Boost install for 1.67 in GTSAM. That adds a good 5 minutes to the CI.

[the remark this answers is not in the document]
dellaert: Tell me again why we do not support the default boost version on 1804?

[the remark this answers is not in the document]
varunagrawal: Because Boost versions before 1.67 have a bug in their serialization where only the object in question is serialized and not its constituent objects. The best analogy is that serialization would do a shallow copy when what we really need is a deep copy.

tests/testRevoluteJoint.cpp
varunagrawal: Yeah look at the number of `using` statements on the left. Does it make sense to have all of them?

.github/workflows/linux-ci.yml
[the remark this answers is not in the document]
ProfFan: I think it's before 1.65 : https://github.com/borglab/gtsam/blob/6f9561eb04c0d930abfdb39e2a18c4f4d5483e3e/INSTALL.md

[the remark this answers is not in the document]
varunagrawal: That doc needs to be updated. If you see the CI for GTSAM, we install 1.67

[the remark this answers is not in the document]
ProfFan: Yes, but I think it works from 1.65. I can make a VM and test.

gtdynamics/universal_robot/Robot.h
```diff
@@ -112,6 +112,32 @@ class Robot {
...
+      // compare the key name and the underlying shared_ptr object
+      return a.first == b.first && (*a.second) == (*b.second);
+    };
+
+    return (this->name_to_link_.size() == other.name_to_link_.size() &&
+            std::equal(this->name_to_link_.begin(), this->name_to_link_.end(),
+                       other.name_to_link_.begin(), link_comparator) &&
+            this->name_to_joint_.size() == other.name_to_joint_.size() &&
+            std::equal(this->name_to_joint_.begin(), this->name_to_joint_.end(),
+                       other.name_to_joint_.begin(), joint_comparator));
```
gchenfc, line 134: note 1: `std::equal` is since c++14
note 2: I think if you give both the begin and end of both ranges and it'll automatically check the size so you don't have to?  apologies if you already tried this and have a reason it won't work

gtdynamics/universal_robot/Robot.h
```diff
@@ -112,6 +112,32 @@ class Robot {
   /// Print links and joints of the robot, for debug purposes
   void print(const std::string &s = "") const;
 
+  /// Overload equality operator.
+  bool operator==(const Robot &other) const {
+    // Define comparators for easy std::map equality checking
+    auto link_comparator = [](decltype(*this->name_to_link_.begin()) a,
+                              decltype(a) b) {
+      // compare the key name and the underlying shared_ptr object
```
gchenfc, line 120: nit: can you \_emph_ or make a note in the comment "Define comparators ..." or something that clarifies that the _only_ reason this is needed is for the `shared_ptr` ?  Spent a couple minutes trying to figure out why you didn't just use the default map comparator 😂

tests/testHelicalJoint.cpp
```diff
@@ -94,6 +98,26 @@ TEST(Joint, params_constructor) {
                       j1->parameters().scalar_limits.value_limit_threshold));
 }
 
+BOOST_CLASS_EXPORT(gtdynamics::HelicalJoint)
+BOOST_CLASS_EXPORT(gtdynamics::RevoluteJoint)
```
gchenfc, line 102: is revolute joint needed here?

gchenfc (approved):
I guess I'll let you and Fan duke it out about Boost, but the rest LGTM

gtdynamics/universal_robot/Robot.h
varunagrawal: 1. `std::equal` is pre-C++11 :slightly_smiling_face: 
2. This version of `std::equal` which takes iterators to the start and end of both ranges is C++14.

gtdynamics/universal_robot/Link.h
```diff
@@ -206,6 +210,34 @@ class Link : public boost::enable_shared_from_this<Link> {
...
+    ar &BOOST_SERIALIZATION_NVP(id_);
+    ar &BOOST_SERIALIZATION_NVP(name_);
+    ar &BOOST_SERIALIZATION_NVP(mass_);
+    ar &BOOST_SERIALIZATION_NVP(centerOfMass_);
+    ar &BOOST_SERIALIZATION_NVP(inertia_);
+    ar &BOOST_SERIALIZATION_NVP(bMcom_);
+    ar &BOOST_SERIALIZATION_NVP(bMlink_);
+    ar &BOOST_SERIALIZATION_NVP(is_fixed_);
+    ar &BOOST_SERIALIZATION_NVP(fixed_pose_);
+    ar &BOOST_SERIALIZATION_NVP(joints_);
```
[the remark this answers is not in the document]
varunagrawal: Yeah I should get rid of joint serialization in Link.h

gtdynamics/universal_robot/Robot.h
varunagrawal: I am still not a fan of this shared pointer implementation of Links and Joints. Other than efficiency (which can also be achieved with references), was there any other reason for choosing to use them?

gtdynamics/universal_robot/Joint.h
```diff
@@ -80,6 +80,20 @@ struct JointParams {
...
+  void serialize(ARCHIVE &ar, const unsigned int /*version*/) {
+    ar &BOOST_SERIALIZATION_NVP(velocity_limit);
+    ar &BOOST_SERIALIZATION_NVP(velocity_limit_threshold);
+    ar &BOOST_SERIALIZATION_NVP(acceleration_limit);
+    ar &BOOST_SERIALIZATION_NVP(acceleration_limit_threshold);
+    ar &BOOST_SERIALIZATION_NVP(torque_limit);
+    ar &BOOST_SERIALIZATION_NVP(torque_limit_threshold);
+    ar &BOOST_SERIALIZATION_NVP(damping_coefficient);
+    ar &BOOST_SERIALIZATION_NVP(spring_coefficient);
+  }
```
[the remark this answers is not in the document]
varunagrawal: Done

.github/workflows/linux-ci.yml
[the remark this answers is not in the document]
varunagrawal: @ProfFan any updates on this?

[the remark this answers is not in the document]
varunagrawal: FYI https://github.com/borglab/gtsam/issues/590#issuecomment-761906003

tests/testHelicalJoint.cpp
varunagrawal: Huh guess not. I don't know why I put it there, guess due to just trying to get the tests to pass.

.github/workflows/linux-ci.yml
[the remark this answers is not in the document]
varunagrawal: Okay, I have a resolution for this: I am just setting Ubuntu 20.04 for the clang CI. That way we are still supporting Boost 1.65 (at least for GCC).

gtdynamics/universal_robot/Robot.h
gchenfc: oh oops my mistake yep thanks :)

gtdynamics/universal_robot/Robot.h
gchenfc: references won't work for polymorphic types, e.g. RevoluteJoint vs ScrewJoint vs PrismaticJoint

specifically, to hold a container of polymorphic types you need pointers (unless use fancy std::variant stuff but I don't think that suits our use case - inherently we need runtime polymorphism to hold all the joints in one data structure)

tests/testHelicalJoint.cpp
gchenfc: no worries

tests/testRobot.cpp
```diff
@@ -231,6 +237,31 @@ TEST(ForwardKinematics, FourBar) {
...
 
+TEST(Robot, Equality) {
+  Robot robot1 = CreateRobotFromFile(
+      kSdfPath + std::string("test/four_bar_linkage_pure.sdf"));
+  Robot robot2 = CreateRobotFromFile(
+      kSdfPath + std::string("test/four_bar_linkage_pure.sdf"));
+
+  EXPECT(robot1 == robot2);
+  EXPECT(robot1.equals(robot2));
+}
```
gchenfc, line 248: sorry I don't want to drag this out longer than it needs to, but could you add a false-y test?  e.g. `robot1.joint1.parent = nullptr; EXPECT(robot1 != robot2)`

I'm personally just a little suspicious of the circular reference serialization thing so want to make sure the round-trip serialization unit tests will cover incorrect serializations (e.g. roundtrip test will always return true if `bool Robot::equals(Robot other) { return true; }`)

gchenfc (approved):
Looks good!

tests/testRobot.cpp
varunagrawal: Good point. Please drag this on as long as needed so this PR is amazing. :slightly_smiling_face:

gtdynamics/universal_robot/Robot.h
varunagrawal: Huh, I figured we could just pass references to the Joint class since the public API for each joint type is the same. I am not going to tackle this anytime soon so we can actually drop it. :sweat_smile:

tests/testRobot.cpp
varunagrawal: Hmm there's no clean way to set the parent link pointer to null. Doing something like this doesn't work either:
```cpp
LinkSharedPtr parent_link = j->parent();
parent_link = nullptr;
```

Also, should this be an equality test or a serialization test? I believe @dellaert mentioned in some other PR or somewhere that either the joints should know about the links or the links should know about the joints and not both. We should probably find that.

gchenfc: I was envisioning an equality test

or maybe something like
`*j->parent() = Link();`
?

Trying to make a false equality, preferably involving pointers.

dellaert:
Serialization should deal fine with pointers, BTW. Not sure what the exact issue is.

tests/testRobot.cpp
varunagrawal: Okay this works. Seems like setting it shared ptr to nullptr doesn't quite work.

.github/workflows/linux-ci.yml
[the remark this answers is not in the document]
varunagrawal: Ping @ProfFan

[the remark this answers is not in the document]
ProfFan: I think this is good, but

.github/workflows/linux-ci.yml
```diff
@@ -17,17 +17,17 @@ jobs:
...
 
         build_type: [Debug, Release]
         include:
-          - name: ubuntu-18.04-gcc-9
-            os: ubuntu-18.04
+          - name: ubuntu-20.04-gcc-9
+            os: ubuntu-20.04
```
ProfFan: They need to be updated^

ProfFan (approved):
Please change gcc back to 18.04 and then LGTM

.github/workflows/linux-ci.yml
varunagrawal: This is so strange. As per commit 1b8707e, that change should have already been made.

PR #318, opened 2021-12-03
Python wrapper update

gchenfc:
I found this cleaning up old local stashes - I think it was needed by Natalie at one point but then wasn't anymore; in any case, it's probably useful to have wrapped still.

@nrakoski3 , if you have time could you make a short unit test for IK drawing from some of your code, as well as tests for any other classes you need from gtdynamics?  I might recommend putting it in `test_forward_kinematics.py` then renaming the file to just `test_kinematics.py`


varunagrawal:
The `PointGoalFactor` might be removed soon. Otherwise this PR looks great.
Can you add some simple unit tests please?

dellaert (approved):

dellaert:
Jerry please add quick unit test before merging. Also merge in master so CI checks that

gchenfc:
will do (after fixing cablerobot segfault)

varunagrawal:
Ping @gchenfc

varunagrawal:
I added the unit tests, merging.

PR #323, opened 2021-12-17
Convert Optimizer Arguments from shared pointers to objects

### LINKED ISSUE #320: Need for shared pointers as arguments?
Great stuff on the new updates.

However I had a question about using shared pointers as function and constructor args. Is there a reason we're doing `Optimizer(const boost::shared_ptr<const OptimizationParameters>& parameters)` instead of `Optimizer(const OptimizationParameters& parameters)`?

Seems like unnecessary verbosity to me especially when `OptimizationParameters` is a simple struct (aka an encapsulation of various parameters).

varunagrawal:
This fixes #320

This PR proposes to swap `Optimizer(const boost::shared_ptr<const OptimizerParams>& params)` with the simpler `Optimizer(const OptimizerParams& params)`. Currently we make use of shared pointers for optimizer parameters which seems unnecessarily verbose since we don't expect to have more than a handful of optimizers running. By using const references, we avoid the need to call `make_shared` at various places and improve the overall readability of the code.

Other benefits include being able to get quick auto-complete with the dot notation, simpler wrapping, and one less dependency on Boost.

### DIFF
--- a/gtdynamics.i
+++ b/gtdynamics.i
@@ -329,8 +329,8 @@ class KinematicsParameters : gtdynamics::OptimizationParameters {
 };
 
 class Kinematics {
-  Kinematics(gtdynamics::KinematicsParameters *parameters =
-                 boost::make_shared<const gtdynamics::KinematicsParameters>());
+  Kinematics(gtdynamics::KinematicsParameters parameters =
+                 gtdynamics::KinematicsParameters());
   gtsam::Values inverse(const gtdynamics::Slice &slice,
                         const gtdynamics::Robot &robot,
                         const gtdynamics::ContactGoals &contact_goals);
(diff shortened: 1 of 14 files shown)


dellaert:
Did you do a design review with Yetong? Does your PR comment give context and rationale for the change? ;-)

dellaert (approved):
that being said, I agree with the change. If Yetong agrees, let’s land. But update PR comment to be more informative, for when we look back at this PR in the future.

varunagrawal:
I was hoping this could serve as a design review. Yetong is more than welcome to shred this PR to pieces. :smile: 
I'll update the PR description though.

yetongumich (approved):
LGTM

PR #325, opened 2021-12-23
Update README and example cmake project

gchenfc:
Previously the cmake projects in `examples` would not compile naively using the instructions in the README.  Specifically, the projects in the `examples` folder cannot be built as standalone cmake projects, they must instead be compiled "under" gtdynamics.

i.e. from GTDynamics/CMakeLists.txt, `add_subdirectory(examples)` then `make example_forward_dynamics.run` works.  However, doing something like

```shell
cd GTDynamics/examples/example_forward_dynamics
mkdir build
cd build
cmake ..
make -j8
```
does __not__ work.

This PR clarifies the README instructions, and also creates an example cmake project that _does_ work standalone.

Note we have a little bit of ugliness with sdformat.  I'll work on making a .cmake file so we can use `find_package(gtdynamics)`.

Also note: I'm not sure what the original intention of the `examples` folder was and how it differs from `scripts` - perhaps these were supposed to be standalone all along but just at some point they stopped working.  Not sure.


varunagrawal:
I think `scripts` should be merged with examples since they both seem to serve the same purpose.

gchenfc:
I think I was thinking `examples` would be more like how to use gtdynamics in external projects (kind of like the gtsam-project-python repo), whereas `scripts` get run inside this repo.  So running `make script` would run the script, but for the examples, those would get standalone outside of the main build directory.

And I guess the `examples` would not be normally built, only the scripts.

But I think more I was just confused why there were both scripts and examples and deduced from the README that this was the intention, so that's why I updated the cmake files to match the README intention.

varunagrawal:
Yeah the scripts directory should probably be merged into the examples directory. Examples are meant to be simple How-Do-I code files.

The scope of this PR should change to represent that.

varunagrawal:
> I think I was thinking `examples` would be more like how to use gtdynamics in external projects (kind of like the gtsam-project-python repo), whereas `scripts` get run inside this repo. So running `make script` would run the script, but for the examples, those would get standalone outside of the main build directory.

The examples to me are more complex than scripts and thus need more files involved. A lot of the examples have the main C++ files along with python files for data generation or visualization. None of those projects really show how to integrate GTD, and the fact that we provide the relevant CMake files makes integration pretty straightforward.

Moreover, I think being able to run `make <specific-example>` and is cleaner so you won't have `CMakeFiles` littered all over in your root directory. That's messy and tedious to clean up.

> And I guess the `examples` would not be normally built, only the scripts.

I don't quite see the reason for this. This is why we have a flag, so users can just turn that flag on when they need the examples built. Having examples within the repo makes more sense since it shows directly how to use certain features.

varunagrawal (approved):
I've updated the project to compile and run, updated the README, and I believe this now looks great. Let's land it.

Issue #324, opened 2021-12-22
Header files installing to root


gchenfc:
Action items:
- @varunagrawal do you have any ideas about this?  If not, this is not urgent or anything but just puzzling.
- @arutkowski could you try
    ```cmake
    message("CMAKE INSTALL INCLUDE DIR is set to:    ${CMAKE_INSTALL_INCLUDEDIR}")
    ```
    at the bottom of `GTDynamics/gtdynamics/CMakeLists.txt` ?

---

@arutkowski is having a strange issue on his machine where cmake is installing the `.h` files into `/gtdynamics/...` (root) instead of `${CMAKE_PREFIX_PATH}/include/gtdynamics/...`.  Interestingly, the `.so` file is properly installed into `${CMAKE_PREFIX_PATH}/lib/libgtdynamics.so`.  Adam worked around it by manually copying the files into the correct directory, but this is not ideal.

Upon reading more carefully, I think maybe this is related to the cmake builtin `CMAKE_INSTALL_INCLUDEDIR` used e.g. here:
https://github.com/borglab/GTDynamics/blob/263658210a4f0fdad9ee8359264667cac5b192e3/gtdynamics/CMakeLists.txt#L43
`CMAKE_INSTALL_INCLUDEDIR` was introduced in [cmake 3.14](https://cmake.org/cmake/help/v3.14/command/install.html) I think, though IIRC @arutkowski  had 3.17 so it should have been there.
Nevertheless, maybe we should do `cmake_minimum_required(VERSION 3.14)`, or just hard-code the includes install path as `include/` instead of using `CMAKE_INSTALL_INCLUDEDIR`.

---

Finally I suspect the use of `CMAKE_INSTALL_INCLUDEDIR` may be inconsistent with
https://github.com/borglab/GTDynamics/blob/263658210a4f0fdad9ee8359264667cac5b192e3/gtdynamics/CMakeLists.txt#L67-L76
which _does_ hardcode the `include/` folder install path.

varunagrawal:
Hardcoding `include/` is not a good idea since that can work differently for different machines. I know for a fact that `include` locations are different on linux vs macos. I would like to resolve that inconsistency directly and remove all hardcoding since that is shoving in assumptions onto cmake which then causes headaches down the line.

Right now, I was going to suggest what @gchenfc said, which is print out the value of `CMAKE_INSTALL_INCLUDEDIR`.
[...]

PR #330, opened 2022-01-12
include(GNUInstallDirs)

yetongumich:
I encountered the same issue as #324, and it turned out `${CMAKE_INSTALL_INCLUDEDIR}` is empty during compilation.

By looking at [CMake documentation](https://cmake.org/cmake/help/latest/command/install.html), I discover that we forgot to include GNUInstallDirs. After adding the line `include(GNUInstallDirs)`, everything works fine.

I'd also suggest to add `make install` into out github workflow so that we can check if installation is successful.

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -86,5 +86,5 @@ jobs:
         working-directory: ./build
       
       - name: Install
-        run: make -j$(nproc) install
+        run: sudo make -j$(nproc) install
         working-directory: ./build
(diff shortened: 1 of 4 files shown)


gchenfc (approved):
Adding `make install` seems reasonable to me as well

yetongumich:
> Adding `make install` seems reasonable to me as well

Sounds good! I'll make another PR to include "make install" in github workflow

varunagrawal:
Just add the make install to this PR. It will be all housekeeping related.

yetongumich:
Added "make install" check and merging

PR #335, opened 2022-02-13
CMake Packaging

### LINKED ISSUE #5: package gtdynamics into a .cmake file so that we can use "find_package" syntax.
This issue was imported from [github.gatech.edu](https://github.gatech.edu/borglab/GTDynamics/issues/16).
The original issue was opened by @aescontrela3 on 2020-03-25T04:43:39Z
---
Refer to #15 for context

varunagrawal:
This PR adds support for making `gtdynamics` a cmake package so we can do `find_package(gtdynamics)` for downstream applications. Fixes #5.

I need this for my own LRSE project and it's been long overdue.
To include gtdynamics, all I needed to do was `find_package(gtdynamics REQUIRED)` and when I print the version as `${gtdynamics_VERSION}`, I get the following output:

```
-- GTSAM include directory:  /usr/local/lib/cmake/GTSAM/../../../include
-- Looking for ignition-math6 -- found version 6.9.2
-- Searching for dependencies of ignition-math6
-- GTSAM Version: 4.2a4
-- GTDynamics Version: 1.0.0
```

This is a CMake update and no C++ changes are made whatsoever. Moreover, it includes commands for configuring and installing mainly 3 files: `gtdynamics-exports.cmake`, `gtdynamicsConfig.cmake` and `gtdynamicsConfigVersion.cmake`. Thus, there should be no breakage in the normal process of installation from source. Even the CI should be evidence of this since we install from source.

Question: Should we stick to `gtdynamics` or convert the package name to `GTDynamics`? The former means we have CMake variables of the form `gtdynamics_VERSION` which is fine IMO, but we can do `GTDynamics_VERSION` instead if you'd like.
GTSAM does `GTSAM_VERSION` but it is also a shorter name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,3 +96,42 @@ message(STATUS "  Cable Robot                               : ${GTDYNAMICS_BUILD
 message(STATUS "  Jumping Robot                             : ${GTDYNAMICS_BUILD_JUMPING_ROBOT}")
 message(STATUS "  Panda Robot                               : ${GTDYNAMICS_BUILD_PANDA_ROBOT}")
 message(STATUS "===============================================================")
+
+# Create the export .cmake file.
+# We should use `install(EXPORT) instead of export()` so it exports the install file.
+install(EXPORT "${PROJECT_NAME}-exports"
+  FILE gtdynamics-exports.cmake
+  DESTINATION lib/cmake/gtdynamics
+)
+
+include(CMakePackageConfigHelpers)
+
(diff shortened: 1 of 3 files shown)


dellaert:
Let’s stick to gtdynamics.
Two questions:
1) how do you know it works? Add evidence in PR comment.
2) add some statement that makes me believe it does not affect anyone negatively
Happy to approve after that - let me know by e-mail.

dellaert (approved):

PR #337, opened 2022-02-19
Fix compile issues

dellaert:
Added missing header.
But, when trying to build wrapper, I do get strange errors:
```
(nbdev) macbook-pro:build dellaert$ make -j12 python-install
[  0%] Generating gtdynamics.cpp
[  4%] Generating jumpingrobot.cpp
[  8%] Generating cablerobot.cpp
No such file or directory
No such file or directoryNo such file or directory

make[3]: *** [python/cablerobot.cpp] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: *** [python/gtdynamics.cpp] Error 1
make[3]: *** [python/jumpingrobot.cpp] Error 1
make[2]: *** [python/CMakeFiles/pybind_wrap_gtdynamics.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....
Consolidate compiler generated dependencies of target gtdynamics
[ 83%] Built target gtdynamics
make[1]: *** [python/CMakeFiles/python-install.dir/rule] Error 2
make: *** [python-install] Error 2
```

Any clues ? @yetongumich @gchenfc ?

PR #338, opened 2022-02-19
Fix to get gtdynamics to compile

varunagrawal:
Recent updates from GTSAM now require us to explicitly include `boost/format.hpp`.

### DIFF
--- a/gtdynamics/utils/JsonSaver.h
+++ b/gtdynamics/utils/JsonSaver.h
@@ -17,6 +17,7 @@
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
 #include <gtsam/slam/PriorFactor.h>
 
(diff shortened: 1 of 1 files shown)


varunagrawal:
Also as an FYI, `JsonSaver.h` has a license referring to Holomatic. We should probably update it or remove it to avoid potential legal issues?

dellaert:
Uhm, #337 ? Also, you should really carefully think about including things like boost::format in headers.

varunagrawal:
Huh didn't see that, thank you! I figured that JsonSaver.h already had a boost import so adding another wouldn't be too bad?

dellaert:
Headers should be as specific as possible, esp. boost headers, because they can substantially increase compile times.

PR #341, opened 2022-03-15
Replace "" include with <>

### LINKED ISSUE #332: [RFC] Convert all includes to <>
Lots of includes are of the form `#include "path to header"`, so we should probably update them to `#include <path to header>` so it is similar to GTSAM and akin to a proper library.

This is a request for a design review. It is really simple to do and it works locally for me (did this back in the first week of December '21).

varunagrawal:
Fixes #332

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -40,6 +40,7 @@ jobs:
             echo "CXX=g++-${{ matrix.version }}" >> $GITHUB_ENV
 
           else
+            sudo apt-get -y clean && sudo apt-get -y update
             sudo apt-get install -y clang-${{ matrix.version }} g++-multilib
             echo "CC=clang-${{ matrix.version }}" >> $GITHUB_ENV
             echo "CXX=clang++-${{ matrix.version }}" >> $GITHUB_ENV
--- a/gtdynamics/cablerobot/factors/CableTensionFactor.cpp
+++ b/gtdynamics/cablerobot/factors/CableTensionFactor.cpp
@@ -9,7 +9,6 @@
 #include "CableTensionFactor.h"
 
 #include <gtdynamics/utils/DynamicsSymbol.h>
-
 #include <gtsam/base/Matrix.h>
 #include <gtsam/base/Vector.h>
 #include <gtsam/geometry/Pose3.h>
--- a/gtdynamics/dynamics/DynamicsGraph.cpp
+++ b/gtdynamics/dynamics/DynamicsGraph.cpp
@@ -11,8 +11,16 @@
  * @author Yetong Zhang, Alejandro Escontrela
  */
 
-#include "gtdynamics/dynamics/DynamicsGraph.h"
-
+#include <gtdynamics/dynamics/DynamicsGraph.h>
+#include <gtdynamics/factors/ContactDynamicsFrictionConeFactor.h>
+#include <gtdynamics/factors/ContactDynamicsMomentFactor.h>
+#include <gtdynamics/factors/ContactHeightFactor.h>
+#include <gtdynamics/factors/ContactKinematicsAccelFactor.h>
+#include <gtdynamics/factors/ContactKinematicsTwistFactor.h>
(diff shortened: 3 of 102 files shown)


gchenfc (commented):
I probably should have brought these questions up in RFC, but anyway:

Although the re-ordering of includes is nice, I'm not clear what the advantage is of using <> instead of <>.  I tried to do some reading but still not 100% clear what (or if) the usual convention is for "a proper library" (as you say).  For example,
* the [Google style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) makes it look like anything in "this project" (which I would take to also mean "this library") should use quotes, and only external projects should use <>
* [cppreference](https://en.cppreference.com/w/cpp/preprocessor/include) says this kind of vague quote:
  > Typical implementations search only standard include directories for syntax (1). The standard C++ library and the standard C library are implicitly included in these standard include directories. 
  
  where syntax (1) is <>.  But this makes me think that only c/c++ std libs should use <> and everything else should use ""?
* Eigen (just the first "proper library" example I happened to think of) uses the "" syntax for its own files
* Boost I wasn't able to recognize a clear pattern, but it appears to use the "" syntax for files in the same "lib", and uses the <> syntax only for files in a different lib (e.g. [boost optional](https://github.com/boostorg/optional/blob/5a444eb84b67a5035e3577476234189c57ca55c1/include/boost/optional/optional_io.hpp))

But maybe considering practical usage would be a better discussion.

One practical issue I would be concerned about is I recall once I had something set up wrong and my code was using the installed version of the include instead of the local version, so when I was changing the local version everything compiled fine but then didn't make the changes I was expecting.  I'm not sure if changing to "" would help at all (since it falls back to <>), but might be more clear it's local?

It seems to me the primary advantage of using <> is that it is less likely to collide with files in projects that include gtdynamics, but I doubt this is much of a concern since all the includes start with "gtdynamics/..." so it doesn't seem like there's any real advantage of using <> ?

varunagrawal:
> I probably should have brought these questions up in RFC, but anyway:

Yes. This is a discussion for that issue (which is why I created it and tagged it as an RFC).

>   where syntax (1) is <>.  But this makes me think that only c/c++ std libs should use <> and everything else should use ""?

Not necessarily.
The use of angle brackets is to tell the compiler/linker where to look for specific headers. If we release GTDynamics as a library, we want the compiler looking in the standard install directories as well and not limiting itself to the project directory. [Relevant SO answer](https://stackoverflow.com/a/49226070/1236990).

> * Eigen (just the first "proper library" example I happened to think of) uses the "" syntax for its own files
> * Boost I wasn't able to recognize a clear pattern, but it appears to use the "" syntax for files in the same "lib", and uses the <> syntax only for files in a different lib (e.g. [boost optional](https://github.com/boostorg/optional/blob/5a444eb84b67a5035e3577476234189c57ca55c1/include/boost/optional/optional_io.hpp))

This seems to be a matter of choice for the developers. Generally if something isn't a part of the external API, it would be fine to include it with "".
 
> One practical issue I would be concerned about is I recall once I had something set up wrong and my code was using the installed version of the include instead of the local version, so when I was changing the local version everything compiled fine but then didn't make the changes I was expecting. I'm not sure if changing to "" would help at all (since it falls back to <>), but might be more clear it's local?

This is a CMake issue where it seems the variables weren't set up correctly.

> It seems to me the primary advantage of using <> is that it is less likely to collide with files in projects that include gtdynamics, but I doubt this is much of a concern since all the includes start with "gtdynamics/..." so it doesn't seem like there's any real advantage of using <> ?

The points you raised are a bit irrelevant because you're considering GTDynamics development itself rather than using it as an external library (I may be wrong but your wording seems to indicate this). If GTDynamics was released by company X, you would want it to install in the correct locations and the compiler to know where to look without specifying specific flags like `-I`. It's the same reason we use <> in GTSAM and not "".

yetongumich (approved):

yetongumich:
I think we shall merge this in.

varunagrawal:
@gchenfc should we merge this in?

PR #343, opened 2022-04-08
WIP: Chains integration in factor graphs

danbarla:
In this PR we incorporate the Chain class and constraints we created earlier in factor graphs.


danbarla:
testChainGraph tests for equality of old graph and new graph.
Can't make it work at the moment.

PR #345, opened 2022-04-12
Improve factors and add some tests

varunagrawal:
This PR improves the `ContactEqualityFactor` and the `JointMeasurementFactor` as well as adds some python tests.

### DIFF
--- a/gtdynamics/factors/JointMeasurementFactor.h
+++ b/gtdynamics/factors/JointMeasurementFactor.h
@@ -34,6 +34,9 @@ class JointMeasurementFactor
   double measured_joint_coordinate_;
 
  public:
+  // shorthand for a smart pointer to a factor
+  using shared_ptr = boost::shared_ptr<JointMeasurementFactor>;
+
   /**
    * @brief Construct a new Link Pose Factor object
    *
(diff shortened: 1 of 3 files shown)


gtdynamics/factors/JointMeasurementFactor.h
```diff
@@ -92,5 +95,21 @@ class JointMeasurementFactor
     gtsam::traits<double>::Print(measured_joint_coordinate_, "  measured: ");
     this->noiseModel_->print("  noise model: ");
   }
+
+  bool equals(const gtsam::NonlinearFactor& other,
+              double tol = 1e-9) const override {
+    const This* e = dynamic_cast<const This*>(&other);
+    return e != nullptr && Base::equals(*e, tol) &&
```
danbarla, line 102: Don't need `e != nullptr` here, just use `e && ...`

gtdynamics/factors/JointMeasurementFactor.h
```diff
@@ -15,6 +15,7 @@
 #pragma once
 #include <gtdynamics/universal_robot/Joint.h>
 #include <gtdynamics/universal_robot/Link.h>
+#include <gtdynamics/utils/values.h>
```
danbarla, line 18: why ? It compiled and passed CI before no?

danbarla (approved):
Looks good to me :)
Small comments below.

gtdynamics/factors/JointMeasurementFactor.h
varunagrawal: True but it's nice to be explicit. :)

varunagrawal:
Thanks Dan!

PR #348, opened 2022-04-14
Update SDFormat to version 12

### LINKED ISSUE #346: Upgrade to SDFormat12
SDFormat10 and SDFormat11 have been deprecated due to their end-of-life. As a result, the CI will fail and we should take steps to upgrade.

varunagrawal:
Fixes #346 

I got a response from the OSRF folks at the issue I raised and based on their recommendation, I updated the SDF file loading logic.


danbarla (approved):
Awesome!

PR #349, opened 2022-04-25
Set default flag to false for preserve_fixed_joint

varunagrawal:
As discussed in today's meeting, setting the `preserve_fixed_joint` flag to false by default.

I added a bunch of unit tests and also wrapped this in Python.


gchenfc (commented):
LGTM, but also IIRC frank asked @danbarla to review so I'll leave him to "approve"

varunagrawal:
Ah shoot you're correct.

danbarla (approved):
LGTM, I believe that's the way it should be.

PR #353, opened 2022-05-16
fix jointIndex syntax error

devarsi-rawal:
Hi all! I noticed there was a small syntax error in sim.py when I was running `python kinematic_mp_sim.py` for the quadruped. I was getting `TypeError: function missing required argument 'bodyIndex' (pos 1)` and noticed that the `bodyindex` param did not have the right capitalization, so here's a quick fix.

### DIFF
--- a/python/gtdynamics/sim.py
+++ b/python/gtdynamics/sim.py
@@ -23,7 +23,7 @@ def set_joint_angles(pyb,
     """
     for joint_id in joint_to_jid_map.values():
         pyb.setJointMotorControl2(bodyUniqueId=robot,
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Thanks !

varunagrawal:
I don't understand something, the PR comment says `bodyIndex` but the change is `jointIndex`. Can someone explain?

devarsi-rawal:
I can see the confusion and I'm honestly not quite sure where `bodyIndex` came from... I think I had handwritten that error message and just made a typo in the PR comment.

PR #356, opened 2022-06-09
Printing Improvements

varunagrawal:
1. Print the build type for CMake.
2. Print the specific Joint type. Now we get info on the exact joint type (e.g. Prismatic, Revolute etc).
3. Print both `bMlink` for a robot as well. I have found this incredibly useful for debugging.


danbarla (approved):
LGTM

PR #292, opened 2021-10-16
Replace `AdjointMapJacobianQ` with `gtsam::Adjoint` w/ Jacobians

gchenfc:
After https://github.com/borglab/gtsam/pull/885 merges, then this will use, e.g.

```c++
Pose3 T = joint->parentTchild(q, T_H_q);
Vector6 twist_p = T.Adjoint(twist_c, H_T, H_twist_c);
H_q = H_T * T_H_q;
```
instead of:
```c++
Vector6 twist_p = T.Adjoint() * twist_c;
H_q = AdjointMapJacobianQ(joint, q) * twist_c;
```

This is just much more sane and generalizeable and uses more GTSAM code instead of custom code.

PR #358, opened 2022-06-24
AdjointMap Jacobians 2

varunagrawal:
This PR updates #292 so we can land it.

Unfortunately, I cannot completely eliminate `AdjointMapJacobianQ` since `Chain::DynamicalEquality3` requires it. One solution would be to move the jacobian computation of `AdjointMap` to GTSAM. This would be a nice add-on.

@danbarla requesting your review since I made some changes to `Chain.cpp`.

### DIFF
--- a/gtdynamics/dynamics/Chain.cpp
+++ b/gtdynamics/dynamics/Chain.cpp
@@ -86,6 +86,35 @@ Pose3 Chain::poe(const Vector &q, boost::optional<Pose3 &> fTe,
   return poe;
 }
 
+/**
+ * Calculate AdjointMap jacobian w.r.t. joint coordinate q
+ * @param q joint angle
+ * @param jMi this COM frame, expressed in next link's COM frame at rest
+ * @param screw_axis screw axis expressed in kth link's COM frame
+ *
+ * TODO(Varun) Perhaps move part of this to gtsam::Pose3::AdjointMap()?
+ */
+gtsam::Matrix6 AdjointMapJacobianQ(double q, const gtsam::Pose3 &jMi,
+                                   const gtsam::Vector6 &screw_axis) {
+  // taking opposite value of screw_axis_ is because
+  // jTi = Pose3::Expmap(-screw_axis_ * q) * jMi;
+  gtsam::Vector3 w = -screw_axis.head<3>();
+  gtsam::Vector3 v = -screw_axis.tail<3>();
+  gtsam::Pose3 kTj = gtsam::Pose3::Expmap(-screw_axis * q) * jMi;
+  auto w_skew = gtsam::skewSymmetric(w);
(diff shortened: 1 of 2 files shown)


gtdynamics/dynamics/Chain.cpp
```diff
@@ -111,20 +140,30 @@ gtsam::Vector3 Chain::DynamicalEquality3(
...
     auto ad_J_angles1 = AdjointMapJacobianQ(angles(1), Pose3(), axes_.col(1));
-    auto ad_J_angles2 = AdjointMapJacobianQ(angles(2), Pose3(), axes_.col(2));
-
-    // Calculate the invers adjoint maps of the Poses, as we do in * operator
-    Pose3 p1 = Pose3::Expmap(axes_.col(1) * angles(1));
-    Pose3 p2 = Pose3::Expmap(axes_.col(2) * angles(2));
-    auto ad_inv_p1 = p1.inverse().AdjointMap();
-    auto ad_inv_p2 = p2.inverse().AdjointMap();
 
     // calculate the non-zero terms
```
danbarla, line 163: For example, this made more sense to me the old way since this is the Jacobian for q, so need to multiply in the end with wrench.

gtdynamics/dynamics/Chain.cpp
```diff
@@ -111,20 +140,30 @@ gtsam::Vector3 Chain::DynamicalEquality3(
     // This means that the 3*3 jacobian has an upper triangular structure.
     Matrix A = gtsam::Z_3x3;
```
danbarla, line 142: Hey Varun,
I understand you moved the adjointMapJacobianQ here, that's fine.
I prefer at this point to leave these jacobian calculations as is.
They are not trivial (at least not for me) and they make sense to me as is right now.
I will eventually get rid of this function.

danbarla (commented):
Hey,
For the chain jacobian calculations, I prefer to leave as is for now.
I will get rid of this function eventually.

gtdynamics/dynamics/Chain.cpp
varunagrawal: Wouldn't it make more sense to just get the Jacobian of the Adjoint directly rather than compute the Adjoint map and then multiply it with the wrench?

Aka

```
p1.Adjoint (wrench, &ad)
```

rather than
```
ad = p1.AdjointMap() * transpose
```


I'm sure it you take a minute to understand this, it will make sense and be even cleaner (I removed a lot of duplicate operations).

varunagrawal:
Okay I undid the changes @danbarla

danbarla (approved):
ok to merge from me

varunagrawal:
Merging since this was pretty much copy-pasta from Gerry's PR.

PR #359, opened 2022-06-26
Better Link printing

varunagrawal:
Improved link printing to help with analysis and understanding.

I also moved some method definitions to the .cpp file, and fixed some docs.

### DIFF
--- a/gtdynamics/universal_robot/Link.cpp
+++ b/gtdynamics/universal_robot/Link.cpp
@@ -10,8 +10,8 @@
  * -------------------------------------------------------------------------- */
 
 /**
- * @file  Joint.cpp
- * @brief Absract representation of a robot joint.
+ * @file  Link.cpp
+ * @brief Abstract representation of a robot link.
  */
 
 #include <gtdynamics/dynamics/Dynamics.h>
@@ -23,6 +23,7 @@
 
 namespace gtdynamics {
 
+/* ************************************************************************* */
 gtsam::Vector6_ Link::wrenchConstraint(
     const std::vector<DynamicsSymbol>& wrench_keys, uint64_t t,
     const boost::optional<gtsam::Vector3>& gravity) const {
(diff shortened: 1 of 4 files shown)


varunagrawal:
@yetongumich this PR is also pretty straightforward. Can you give it a look please? :)

yetongumich:
> @yetongumich this PR is also pretty straightforward. Can you give it a look please? :)

Sure, will review it tonight!

gtdynamics/universal_robot/Link.cpp
```diff
@@ -70,4 +71,40 @@ gtsam::Vector6_ Link::wrenchConstraint(
...
+}
+
+/* ************************************************************************* */
+std::ostream& operator<<(std::ostream& os, const Link& link) {
+  std::string fixed = link.isFixed() ? " (fixed)" : "";
+  os << link.name() << ", id=" << size_t(link.id()) << fixed << ":\n";
+  os << "\tcom pose:  " << link.bMcom().rotation().rpy().transpose() << ", "
+     << link.bMcom().translation().transpose() << "\n";
+  os << "\tlink pose: " << link.bMlink().rotation().rpy().transpose() << ", "
+     << link.bMlink().translation().transpose() << std::endl;
```
yetongumich, line 101: Is there a reason to use `std::endl` to replace `\n`? If so, shall it be applied to all three lines?

yetongumich (approved):
LGTM.

Though, it seems the printout contents are not changed?

gtdynamics/universal_robot/Link.cpp
varunagrawal: I prefer using `std::endl` at the end since it flushes the output sequence to the output buffer, ensuring the output is printed out at that point and not interleaved with other output. This [SO answer](https://stackoverflow.com/a/214076/1236990) provides a good example. You may see output interleaving when calling python code and using both python `print` and doing `link.print()`.

Of course, flushing every time is expensive, so I only do it once at the end rather than 3 times. Hopefully that answers this design choice?

varunagrawal:
@yetongumich yes I ensured the printing was still the same so it wouldn't affect anyone(?). If you have suggestions for printing improvement, I am all ears.

gtdynamics/universal_robot/Link.cpp
yetongumich: Thanks! This also helps clear some of my confusions regarding print statements.

PR #362, opened 2022-08-06
Use homebrew to install GTSAM

varunagrawal:
Use homebrew so that GTSAM installation is potentially faster. This should also allow us to use pre-built binaries in the future.

### DIFF
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -46,12 +46,6 @@ jobs:
           brew tap osrf/simulation
           brew install sdformat12
 
-          # Install CppUnitLite.
-          git clone https://github.com/borglab/CppUnitLite.git
-          cd CppUnitLite && mkdir build && cd $_
-          cmake .. && sudo make -j4 install
-          cd ../../
-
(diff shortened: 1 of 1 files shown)


gchenfc (approved):
This comment is probably best left on the homebrew-core repo, but I think the homebrew formula needs some work to work outside of ci, for example,

* On Mac, it requires full xcode, but not sure why since I only have xcode CLI and always compile gtsam just fine
* On a linux machine I tested on, it said 
    > Warning: No available formula with the name "g++" (dependency of borglab/core/gtsam_develop).
==> Searching for similarly named formulae...

    But the machine has g++ installed so idk I guess brew doesn't know about it or something.

But I guess since this PR specifically is for GTDynamics CI and it works fine on the CI, then this can be approved, but ya, I think the homebrew formula needs some work

varunagrawal:
I'm still learning about homebrew formulae myself, so definitely can be improved. The Formula Cookbook says we should specify dependencies so brew can install them if needed, but in our case, we already have the dependencies installed. I'll see what can be done.

PR #363, opened 2022-08-26
Manifold optimization on cable robot kinodynamics problem

gchenfc:
TODO: describe this PR and figure out a way to separate this from #331


yetongumich:
It seems there is some new stuff added. Would you like to talk about it in person?

yetongumich:
Just realized it is an outdated PR. Will try to clean it up.

PR #365, opened 2022-09-10
Optional parent and child link keys for `poseConstraint`

varunagrawal:
This will allow us to specify custom keys as passed to the second constructor of `PoseFactor`.

### DIFF
--- a/gtdynamics/factors/PoseFactor.h
+++ b/gtdynamics/factors/PoseFactor.h
@@ -59,17 +59,16 @@ inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
  *
  * @param wTp_key Key for parent link's CoM pose in world frame.
  * @param wTc_key Key for child link's CoM pose in world frame.
- * @param q_key Key for joint value.
  * @param cost_model The noise model for this factor.
  * @param joint The joint connecting the two poses
  */
 inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
-    DynamicsSymbol wTp_key, DynamicsSymbol wTc_key, DynamicsSymbol q_key,
+    DynamicsSymbol wTp_key, DynamicsSymbol wTc_key,
     const gtsam::noiseModel::Base::shared_ptr &cost_model,
     JointConstSharedPtr joint) {
   return boost::make_shared<gtsam::ExpressionFactor<gtsam::Vector6>>(
       cost_model, gtsam::Vector6::Zero(),
-      joint->poseConstraint(wTp_key.time()));
+      joint->poseConstraint(wTp_key.time(), wTp_key, wTc_key));
 }
 
 }  // namespace gtdynamics
--- a/gtdynamics/factors/PoseFactor.h
+++ b/gtdynamics/factors/PoseFactor.h
@@ -63,7 +63,7 @@ inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
  * @param joint The joint connecting the two poses
  */
 inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
-    DynamicsSymbol wTp_key, DynamicsSymbol wTc_key,
+    DynamicsSymbol wTp_key, DynamicsSymbol wTc_key, DynamicsSymbol q_key,
     const gtsam::noiseModel::Base::shared_ptr &cost_model,
     JointConstSharedPtr joint) {
   return boost::make_shared<gtsam::ExpressionFactor<gtsam::Vector6>>(
--- a/gtdynamics/factors/PoseFactor.h
+++ b/gtdynamics/factors/PoseFactor.h
@@ -68,7 +68,7 @@ inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
     JointConstSharedPtr joint) {
   return boost::make_shared<gtsam::ExpressionFactor<gtsam::Vector6>>(
       cost_model, gtsam::Vector6::Zero(),
-      joint->poseConstraint(wTp_key.time(), wTp_key, wTc_key));
+      joint->poseConstraint(wTp_key, wTc_key, q_key));
 }
 
 }  // namespace gtdynamics
--- a/gtdynamics/universal_robot/Joint.h
+++ b/gtdynamics/universal_robot/Joint.h
@@ -416,9 +416,16 @@ class Joint : public boost::enable_shared_from_this<Joint> {
    * @brief Create expression that constraint the pose of two links imposed by
    * the joint angle.
    */
-  gtsam::Vector6_ poseConstraint(
-      uint64_t t = 0, const boost::optional<DynamicsSymbol> &wTp = boost::none,
-      const boost::optional<DynamicsSymbol> &wTc = boost::none) const;
+  gtsam::Vector6_ poseConstraint(uint64_t t = 0) const;
+
+  /**
+   * @brief Create expression that constraint the pose of two links imposed by
+   * the joint angle. Alternative version using custom keys.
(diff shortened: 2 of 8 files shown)


gtdynamics/universal_robot/Joint.h
```diff
@@ -416,7 +416,9 @@ class Joint : public boost::enable_shared_from_this<Joint> {
    * @brief Create expression that constraint the pose of two links imposed by
    * the joint angle.
    */
-  gtsam::Vector6_ poseConstraint(uint64_t t = 0) const;
+  gtsam::Vector6_ poseConstraint(
+      uint64_t t = 0, const boost::optional<DynamicsSymbol> &wTp = boost::none,
+      const boost::optional<DynamicsSymbol> &wTc = boost::none) const;
```
gchenfc, line 421: I think this might work better as an overloaded function since supplying t vs (wTp, wTc) should, I think, be mutually exclusive.

And presumably the non-t version should also be forced to specify the joint angle q.  Then that would revert the changes to `DynamicsGraph`, `KinematicsSlice`, PoseFactor line 66, and testPoseFactor.

gtdynamics/factors/PoseFactor.h
```diff
@@ -59,17 +59,16 @@ inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
...
 inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
-    DynamicsSymbol wTp_key, DynamicsSymbol wTc_key, DynamicsSymbol q_key,
+    DynamicsSymbol wTp_key, DynamicsSymbol wTc_key,
     const gtsam::noiseModel::Base::shared_ptr &cost_model,
     JointConstSharedPtr joint) {
   return boost::make_shared<gtsam::ExpressionFactor<gtsam::Vector6>>(
       cost_model, gtsam::Vector6::Zero(),
-      joint->poseConstraint(wTp_key.time()));
+      joint->poseConstraint(wTp_key.time(), wTp_key, wTc_key));
 }
```
gchenfc, line 72: like `joint::poseConstraint`, maybe this constructor should be overloaded instead, so one constructor is like the previous version, and the other constructor is like

```c++
inline gtsam::NoiseModelFactor::shared_ptr PoseFactor(
    uint64_t time,
    const gtsam::noiseModel::Base::shared_ptr &cost_model,
    JointConstSharedPtr joint) {
  ...
}
```

gchenfc (changes requested):
I think that supplying `time` should be mutually exclusive with supplying the 3 keys, so I think these should be overloads instead of optional args.

Also, more generally, it seems odd that there should be cases when you are using a joint's pose calculations but not using the same keys as the parent/child links (some inconsistency), but maybe this is because some other code has previously used certain keys for links and you don't want to interfere with that or something?

varunagrawal:
> Also, more generally, it seems odd that there should be cases when you are using a joint's pose calculations but not using the same keys as the parent/child links (some inconsistency), but maybe this is because some other code has previously used certain keys for links and you don't want to interfere with that or something?

This is because for the base link of a robot I am using a GTSAM key `X(t)` and not a GTD key, so the key value (as an int) doesn't match what `PoseKey(base_link->id(), t)` will return. I think of keys as symbolic pointers (like a person's name) so I can ask the joint for a result but I can refer to its links by a different names right?

I also decided to add this since it allows for greater flexibility. E.g. GPMP2 was written before GTD but could benefit from GTD. I can now use GTD features but keep changes to GPMP2 to a minimum, especially for someone like Matt who knows GTSAM well but isn't acquainted with the key structure (aka `DynamicsSymbol`) of GTD.

gtdynamics/factors/PoseFactor.h
varunagrawal: This constructor already exists and is the "canonical" one. The one I updated in this PR is the overloaded one.

Maybe I am not understanding something about your comment?

gchenfc: oh ya my bad, I didn't see that

gtdynamics/universal_robot/Joint.h
varunagrawal: What is the reasoning for the constructors to be mutually exclusive? There would be code duplication in the overloaded method since the key is needed for the expression and I wanted to minimize the change.

Also going back to your prior comment about keys of the joint's link, that is something I found strange here. Why are we specifying the joint key when we have the joint? I can add the key back but then we would have to add a similar check for the joint key and use the specified one if it is not null.

varunagrawal: Correction: I will add the JointAngleKey back. :)

gchenfc: Re: code duplication, I was thinking something like:

```c++
Vector6_ poseConstraint(const DynamicsSymbol &wTp,
                        const DynamicsSymbol &wTc,
                        const DynamicsSymbol &q) const {
  ...
}
Vector6_ poseConstraint(uint64_t t) const {
  return poseConstraint(PoseKey(parent->id(), t), PoseKey(child->id(), t), JointAngleKey(id(), t));
}
```

As for reasoning, it seemed weird to me to mix-and-match custom vs default keys.  As it stands, you can't specify a custom joint angle which seems inconsistent with the nature of the change, and if you could, then there are cases where the t argument is completely unused (i.e. when all 3 custom keys are specified).  But idk I guess 

>Also going back to your prior comment about keys of the joint's link, that is something I found strange here. Why are we specifying the joint key when we have the joint? I can add the key back but then we would have to add a similar check for the joint key and use the specified one if it is not null.

I'm not sure I follow this, but if you're referring to why PoseFactor constructor originally had 3 keys passed in, I am confused by that (as well).  If you're referring to why _I_ think the joint angle q should also be specified, I'd say for the same reason that you want the link pose keys specified.  "why are we specifying the joint key when we have the joint?" can be stated the same way for the pose key: "why are we specifying the link pose key when we have the joint (which has pointers to the links)?"  Mainly I guess I'm just thinking to be consistent.

varunagrawal:
This was accidentally closed. :(

@yetongumich

gtdynamics/universal_robot/Joint.h
yetongumich: I agree with Gerry's way of redesigning the function signatures.

But let's merge this first since it's related to the code on quadruped state estimation project, and I'll make a future PR to address this.

yetongumich (approved):

gtdynamics/universal_robot/Joint.h
varunagrawal: > As it stands, you can't specify a custom joint angle which seems inconsistent with the nature of the change, and if you could, then there are cases where the t argument is completely unused (i.e. when all 3 custom keys are specified). But idk I guess

I am not sure I follow. I thought the idea was to specify a custom joint angle via a prior on the joint angle variable. The thought behind this change was that we always set the joint angle variable to be a function of the joint ID and the time `t` hence providing a custom joint angle key didn't make sense?

> I'm not sure I follow this, but if you're referring to why PoseFactor constructor originally had 3 keys passed in, I am confused by that (as well). If you're referring to why I think the joint angle q should also be specified, I'd say for the same reason that you want the link pose keys specified. "why are we specifying the joint key when we have the joint?" can be stated the same way for the pose key: "why are we specifying the link pose key when we have the joint (which has pointers to the links)?" Mainly I guess I'm just thinking to be consistent.

I see your point about consistency here. Changing the link key made sense since link's are the primary objects that have mass and pose so they are the ones we care about.

gchenfc (approved):

PR #367, opened 2022-09-20
Key Helpers return `gtsam::Key`

varunagrawal:
Something that has been a bit irksome for me has been the need to specify `.key()` for the various key helpers.

Example:
```python
j_key = gtd.JointAngleKey(0, t=10).key()
```

instead of
```python
j_key = gtd.JointAngleKey(0, t=10)
```
which is both cleaner and shorter. If the function says "Key" it should return a Key, right?

This PR updates the code to remedy this. Updating dependent code should be a cinch (global replace `.key()` with the empty string) and should make the library more user friendly.

### DIFF
--- a/gtdynamics/jumpingrobot/src/jr_visualizer.py
+++ b/gtdynamics/jumpingrobot/src/jr_visualizer.py
@@ -43,10 +43,10 @@ def update_jr_frame(ax, values, jr, k):
         z = pose.z()
         theta = pose.rotation().roll()
         l = 0.55
-        start_y = y - l/2 * np.cos(theta)
-        start_z = z - l/2 * np.sin(theta)
-        end_y = y + l/2 * np.cos(theta)
-        end_z = z + l/2 * np.sin(theta)
+        start_y = y - l / 2 * np.cos(theta)
+        start_z = z - l / 2 * np.sin(theta)
+        end_y = y + l / 2 * np.cos(theta)
+        end_z = z + l / 2 * np.sin(theta)
 
         ax.plot([start_y, end_y], [start_z, end_z], color=color)
 
@@ -92,6 +92,7 @@ def visualize_jr_trajectory(values, jr, num_steps, step=1):
 
     def animate(i):
         update_jr_frame(ax, values, jr, i)
+
     frames = np.arange(0, num_steps, step)
     FuncAnimation(fig, animate, frames=frames, interval=10)
     plt.show()
@@ -100,11 +101,13 @@ def visualize_jr_trajectory(values, jr, num_steps, step=1):
 def make_plot(values, jr, num_steps):
     """ Draw plots of all quantities with time. """
     joint_names = ["knee_r", "hip_r", "hip_l", "knee_l"]
-    colors = {"knee_r": "red",
-              "hip_r": "orange",
-              "hip_l": "green",
-              "knee_l": "blue",
-              "source": "black"}
+    colors = {
+        "knee_r": "red",
+        "hip_r": "orange",
+        "hip_l": "green",
+        "knee_l": "blue",
+        "source": "black"
+    }
 
     qs_dict = {name: [] for name in joint_names}
     vs_dict = {name: [] for name in joint_names}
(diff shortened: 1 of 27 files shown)


varunagrawal:
For debugging in C++, it should be straightforward to convert the Key back to a DynamicsSymbol (I believe there is a dicrect constructor for that). Plus the GTDKeyFormatter already handles this well if you prefer print debugging.

I am not sure I follow your comment on `implicitly_convertible`. I may be missing some context?

varunagrawal:
Also, from personal experience, I've never had to extract the keys and convert them. Rather I just use the GTDKeyFormatter and get the same debugging info. This works well since I use a mix of GTSAM and GTD keys and the key formatter converts them all to the correct GTD format. 🙂

gchenfc:
> For debugging in C++, it should be straightforward to convert the Key back to a DynamicsSymbol (I believe there is a dicrect constructor for that). Plus the GTDKeyFormatter already handles this well if you prefer print debugging.

I would argue that:  
`print(gtd.DynamicsSymbol(key))`  
or  
`print(gtd.GTDKeyFormatter(key))`  
are more annoying than  
`key.key()`

> I am not sure I follow your comment on `implicitly_convertible`. I may be missing some context?

https://pybind11.readthedocs.io/en/stable/advanced/classes.html#implicit-conversions

This is I think the exact use case we have, where we want `Key` to be implicitly convertible to `DynamicsSymbol`.  But this may not be possible since Key is typedef'd to a native type (not pybind object).  But might be worth looking into.

> Also, from personal experience, I've never had to extract the keys and convert them. Rather I just use the GTDKeyFormatter and get the same debugging info. This works well since I use a mix of GTSAM and GTD keys and the key formatter converts them all to the correct GTD format. 🙂

From my personal experience, I have ;) and do use the debugging info enough that I prefer not to have to type `GTDKeyFormatter` if possible ;)

More generally speaking, this PR is "erasing type information" which I don't think is a great design choice?

---

My personal opinion: if we can get implicitly_convertible working, we should do that.  Otherwise, I think the annoyance of typing `.key()` is worth having normal DynamicsSymbol usage.  Or perhaps an alternative: rename `JointAngleKey` to `JointAngleSymbol` or something, and have a separate `JointAngleKey` that returns the .key() version.  But I'm not sure how I feel about that either.

gchenfc:
FYI https://github.com/pybind/pybind11/pull/264

varunagrawal:
I don't understand either of the asks.

The implicit conversion is already happening at the cpp level so why do we need to do it at the python wrapper level again?

Also, I don't believe you have to call DynamicsSymbol on any key. Rather you should be able to pass DynamicsSymbol as a key directly or retrieve a key as DynamicsSymbol and the compiler will take care of the necessary conversion.

This was the primary reason for this PR. When the conversion is implicitly happening, there should be no reason for us to call .key separately.

gchenfc:
Varun and I chatted offline - here's the summary/conclusion:

The closest analog to these `JointAngleKey`, `JointVelKey`, ... functions is `gtsam::symbol_shorthand`, and those return `Key`s, i.e. `inline Key X(int i) { return Symbol('X', i); }`, so it makes the most sense for these functions to follow the same pattern.  Therefore this PR makes sense.  i.e.

```c++
inline gtsam::Key JointAngleKey(args) { return DynamicsSymbol(args); }
```

Varun will make a small commit and I'll review shortly thereafter.

gtdynamics/jumpingrobot/src/jr_visualizer.py
```diff
@@ -124,39 +127,62 @@ def make_plot(values, jr, num_steps):
...
-            pressures_dict[name].append(values.atDouble(Actuator.PressureKey(j, k)))
+            pressures_dict[name].append(
+                values.atDouble(Actuator.PressureKey(j, k)))
             masses_dict[name].append(values.atDouble(Actuator.MassKey(j, k)))
-            mdots_dict[name].append(values.atDouble(Actuator.MassRateActualKey(j, k)))
-            contractions_dict[name].append(values.atDouble(Actuator.ContractionKey(j, k)))
+            mdots_dict[name].append(
+                values.atDouble(Actuator.MassRateActualKey(j, k)))
+            contractions_dict[name].append(
+                values.atDouble(Actuator.ContractionKey(j, k)))
```
gchenfc, line 136: Some of the changes in this file look like just different line lengths in linter/autoformatter settings, and I think the one-line versions are cleaner and not that long, maybe allow what I assume is the 100-char width in this file?

gtdynamics/jumpingrobot/src/jumping_robot.py
```diff
@@ -181,36 +183,42 @@ def create_robot(params, phase) -> gtd.Robot:
...
-        shank_l = JumpingRobot.construct_link(
-            5, "shank_l", mass_list[0], length_list[0], link_radius, link_poses["shank_l"])
+        shank_r = JumpingRobot.construct_link(1, "shank_r", mass_list[4],
+                                              length_list[4], link_radius,
+                                              link_poses["shank_r"])
+        thigh_r = JumpingRobot.construct_link(2, "thigh_r", mass_list[3],
+                                              length_list[3], link_radius,
+                                              link_poses["thigh_r"])
+        torso = JumpingRobot.construct_link(3, "torso", mass_list[2],
+                                            length_list[2], link_radius,
```
gchenfc, line 198: same comment as for the previous file - it also doesn't look like either of these files actually have any change involving the `.key` so I don't think they belong here.

gtdynamics/utils/values.h
```diff
@@ -39,45 +39,60 @@ class KeyDoesNotExist : public gtsam::ValuesKeyDoesNotExist {
...
+/* Shorthand for dt_k, for duration for timestep dt_k during phase k. */
+inline gtsam::Key PhaseKey(int k) {
+  return DynamicsSymbol::SimpleSymbol("dt", k);
+}
+
+/* Shorthand for t_k, time at time step k. */
+inline gtsam::Key TimeKey(int k) {
+  return DynamicsSymbol::SimpleSymbol("t", k);
+}
+
```
gchenfc, line 95: I didn't notice yetong's key functions changing, did you do those too?  e.g. `Actuator.MassKey`, ...

not sure where those would be...

gchenfc (approved):
LGTM - I left a couple comments about Yetong's code (not needing to auto-format files that appear otherwise untouched by this PR, and whether some other key generator functions from jumping robot were updated or not)

gtdynamics/jumpingrobot/src/jumping_robot.py
varunagrawal: Yeah I thought formatting these files made sense now since the main changes in this PR are quite small. I am surprised these were not formatted earlier. 😕

gtdynamics/jumpingrobot/src/jr_visualizer.py
varunagrawal: Is line-length=100 the default line length we accept? This is the same setting as what the other python files are formatted with, so in the interest of consistency, I am genuinely wondering what the correct approach is here.

gtdynamics/utils/values.h
varunagrawal: Those are in `jumping_robot.py` and already call `.key` on a Symbol object, so that looks good.

gtdynamics/jumpingrobot/src/jumping_robot.py
yetongumich: I think I may have used a different format option before when writing these files.

gtdynamics/utils/values.h
yetongumich: That shall not be a problem

yetongumich (approved):

gtdynamics/jumpingrobot/src/jr_visualizer.py
gchenfc: oh I thought maybe you set to 80-char, if this is 100 then nvm :)

PR #368, opened 2023-01-24
Fixes as per GTSAM 4.2

varunagrawal:
Updates to get GTD to compile with GTSAM 4.2.

### DIFF
--- a/gtdynamics/utils/Initializer.cpp
+++ b/gtdynamics/utils/Initializer.cpp
@@ -37,15 +37,15 @@ using gtsam::Vector6;
 
 namespace gtdynamics {
 
-Pose3 Initializer::AddGaussianNoiseToPose(const Pose3& T, const Sampler& sampler) const {
+Pose3 Initializer::AddGaussianNoiseToPose(const Pose3& T,
+                                          const Sampler& sampler) const {
   Vector6 xi = sampler.sample();
   return T.expmap(xi);
 }
 
-std::vector<Pose3> Initializer::InterpolatePoses(const Pose3& wTl_i,
-                                    const std::vector<Pose3>& wTl_t, double t_i,
-                                    const std::vector<double>& timesteps,
-                                    double dt) {
+std::vector<Pose3> Initializer::InterpolatePoses(
+    const Pose3& wTl_i, const std::vector<Pose3>& wTl_t, double t_i,
+    const std::vector<double>& timesteps, double dt) {
   std::vector<Pose3> wTl_dt;
   Pose3 wTl = wTl_i;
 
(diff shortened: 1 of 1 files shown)


varunagrawal:
Looks like we need a PPA of the 4.2 release. @dellaert @ProfFan do we have that?

dellaert:
> Looks like we need a PPA of the 4.2 release. @dellaert @ProfFan do we have that?

Not finalized yet. When I finalize (actually make a tag and verify the wrappers) I'll need help cutting a PPA. 4.2a8 is the latest "official" release at this point. The other is a release *branch*.

gtdynamics/utils/Initializer.cpp
```diff
@@ -145,8 +144,11 @@ Values Initializer::InitializeSolutionInterpolation(
     InsertTwist(&init_vals, link->id(), t, gtsam::Z_6x1);
     init_vals = robot.forwardKinematics(init_vals, t, link_name);
 
-    for (auto&& kvp : ZeroValues(robot, t, gaussian_noise, contact_points)) {
-      init_vals.tryInsert(kvp.key, kvp.value);
+    Values zero_values = ZeroValues(robot, t, gaussian_noise, contact_points);
+    for (auto&& key : zero_values.keys()) {
```
dellaert, line 148: This works, but I think this is even better:
init_vals.insert(zero_values);

gtdynamics/utils/Initializer.cpp
```diff
@@ -145,8 +144,11 @@ Values Initializer::InitializeSolutionInterpolation(
     InsertTwist(&init_vals, link->id(), t, gtsam::Z_6x1);
     init_vals = robot.forwardKinematics(init_vals, t, link_name);
 
-    for (auto&& kvp : ZeroValues(robot, t, gaussian_noise, contact_points)) {
-      init_vals.tryInsert(kvp.key, kvp.value);
+    Values zero_values = ZeroValues(robot, t, gaussian_noise, contact_points);
```
dellaert, line 147: const

gtdynamics/utils/Initializer.cpp
```diff
@@ -168,8 +170,10 @@ Values Initializer::InitializeSolutionInterpolationMultiPhase(
         robot, link_name, pose, wTl_t[i], curr_t, ts[i], dt, gaussian_noise,
         contact_points);
 
-    for (auto&& key_value_pair : phase_vals) {
-      init_vals.tryInsert(key_value_pair.key, key_value_pair.value);
+    for (auto&& key : phase_vals.keys()) {
```
dellaert, line 173: init_vals.insert_or_assign(phase_vals);

gtdynamics/utils/Initializer.cpp
```diff
@@ -330,15 +335,18 @@ Values Initializer::MultiPhaseInverseKinematicsTrajectory(
       MultiPhaseZeroValuesTrajectory(robot, phase_steps, transition_graph_init,
                                      dt, gaussian_noise, phase_contact_points);
 
-  for (auto&& key_value_pair : zero_values) {
-    init_vals.tryInsert(key_value_pair.key, key_value_pair.value);
+  for (auto&& key : zero_values.keys()) {
```
dellaert, line 338: insert

dellaert (approved):
Please try my suggestions as that is a better pattern to copy/paste.

gtdynamics/utils/Initializer.cpp
varunagrawal: So that won't work, because the code expects only insertion and not updates. If a key already exists, the expectation is to simply skip it.

This pattern is very weird to me (I didn't write this) so I am not trying to make too many changes.

gtdynamics/utils/Initializer.cpp
varunagrawal: Same deal. `insert_or_assign` doesn't do the same thing as `tryInsert`. 😢

varunagrawal:
I'll add the `const` when updating the CI for the new PPA.

PR #369, opened 2023-01-28
Add support for TBB

varunagrawal:
Added TBB support following the paradigm of GTSAM.

The `FindTBB.cmake` is required since the one provided by TBB doesn't work quite well (you get an error that the lib file can't be found).

This is needed if you want to compile GTSAM with TBB and then link GTD against it.


dellaert:
CI fails?

dellaert (approved):
LGTM, but I’m not the cmake guru

PR #370, opened 2023-02-10
Upgrade GTDynamics to GTSAM 4.3

varunagrawal:
- [x] Update source code.
- [x] Get tests to pass.
- [x] Make Boost serialization a CMake flag.
- [x] Remove boost headers.
- [x] Update examples & scripts.
- [x] Fix wrapper.

### DIFF
--- a/gtdynamics/dynamics/Chain.cpp
+++ b/gtdynamics/dynamics/Chain.cpp
@@ -45,7 +45,7 @@ Chain Chain::compose(std::vector<Chain> &chains) {
   return total_chain;
 }
 
-Pose3 Chain::poe(const Vector &q, boost::optional<Pose3 &> fTe,
+Pose3 Chain::poe(const Vector &q, std::optional<Pose3> fTe,
                  gtsam::OptionalJacobian<-1, -1> J) {
   // Check that input has good size
   if (q.size() != length()) {
@@ -121,7 +121,7 @@ gtsam::Vector3 Chain::DynamicalEquality3(
     gtsam::OptionalJacobian<3, 3> H_angles,
     gtsam::OptionalJacobian<3, 3> H_torques) {
   Matrix J;
-  poe(angles, boost::none, J);
+  poe(angles, {}, J);
   if (H_wrench) {
     // derivative of difference with respect to wrench
     *H_wrench = J.transpose();
--- a/gtdynamics/factors/CollocationFactors.h
+++ b/gtdynamics/factors/CollocationFactors.h
@@ -30,10 +30,10 @@ namespace gtdynamics {
  *
  * @return pose_t1 link pose at next time step
  */
-gtsam::Pose3 predictPose(
-    const gtsam::Pose3 &pose_t0, const gtsam::Vector6 &twistdt,
-    gtsam::OptionalJacobian<6, 6> H_pose_t0 = boost::none,
-    gtsam::OptionalJacobian<6, 6> H_twistdt = boost::none) {
+gtsam::Pose3 predictPose(const gtsam::Pose3 &pose_t0,
+                         const gtsam::Vector6 &twistdt,
+                         gtsam::OptionalJacobian<6, 6> H_pose_t0 = {},
+                         gtsam::OptionalJacobian<6, 6> H_twistdt = {}) {
   gtsam::Matrix6 Hexp;
   gtsam::Pose3 t1Tt0 = gtsam::Pose3::Expmap(twistdt, Hexp);
 
@@ -47,8 +47,8 @@ gtsam::Pose3 predictPose(
 }
 
 /**
- * EulerPoseCollocationFactor is a four-way nonlinear factor between link pose of
- * current and next time steps
+ * EulerPoseCollocationFactor is a four-way nonlinear factor between link pose
+ * of current and next time steps
  */
 class EulerPoseCollocationFactor
     : public gtsam::NoiseModelFactor4<gtsam::Pose3, gtsam::Pose3,
@@ -59,9 +59,9 @@ class EulerPoseCollocationFactor
                                         gtsam::Vector6, double>;
 
  public:
-  EulerPoseCollocationFactor(gtsam::Key pose_t0_key, gtsam::Key pose_t1_key,
-                       gtsam::Key twist_key, gtsam::Key dt_key,
-                       const gtsam::noiseModel::Base::shared_ptr &cost_model)
+  EulerPoseCollocationFactor(
+      gtsam::Key pose_t0_key, gtsam::Key pose_t1_key, gtsam::Key twist_key,
+      gtsam::Key dt_key, const gtsam::noiseModel::Base::shared_ptr &cost_model)
       : Base(cost_model, pose_t0_key, pose_t1_key, twist_key, dt_key) {}
 
   virtual ~EulerPoseCollocationFactor() {}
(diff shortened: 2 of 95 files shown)


varunagrawal:
@gchenfc @yetongumich I haven't updated the cable robot and the jumping robot. Can you make a PR against this PR to address that?

gtdynamics/factors/CollocationFactors.h
```diff
@@ -76,10 +76,10 @@ class EulerPoseCollocationFactor
   gtsam::Vector evaluateError(
       const gtsam::Pose3 &pose_t0, const gtsam::Pose3 &pose_t1,
       const gtsam::Vector6 &twist, const double &dt,
-      boost::optional<gtsam::Matrix &> H_pose_t0 = boost::none,
-      boost::optional<gtsam::Matrix &> H_pose_t1 = boost::none,
-      boost::optional<gtsam::Matrix &> H_twist = boost::none,
-      boost::optional<gtsam::Matrix &> H_dt = boost::none) const override {
+      gtsam::OptionalMatrixType H_pose_t0 = nullptr,
```
gchenfc, line 79: this is an instance where, for consistency, it might make sense to replace all boost::none with `nullptr` or `std::nullopt_t` instead of {}

gtdynamics/factors/ContactEqualityFactor.h
```diff
@@ -66,16 +66,16 @@ class ContactEqualityFactor
...
   /// Generic method to compute difference between contact points and provide
   /// jacobians.
-  gtsam::Vector3 contactPointsDifference(
-      const gtsam::Pose3 &wT1, const gtsam::Pose3 &wT2,
-      boost::optional<gtsam::Matrix &> H1 = boost::none,
-      boost::optional<gtsam::Matrix &> H2 = boost::none) const {
+  gtsam::Vector3 contactPointsDifference(const gtsam::Pose3 &wT1,
+                                         const gtsam::Pose3 &wT2,
+                                         gtsam::OptionalMatrixType H1,
+                                         gtsam::OptionalMatrixType H2) const {
```
gchenfc, line 78: Might have forgotten default args here?

gtdynamics/statics/StaticWrenchFactor.cpp
```diff
@@ -33,13 +33,13 @@ namespace gtdynamics {
...
     const gtsam::noiseModel::Base::shared_ptr &cost_model, double mass,
-    const boost::optional<gtsam::Vector3> &gravity)
+    const std::optional<gtsam::Vector3> &gravity)
     : Base(cost_model, wrench_keys), mass_(mass), gravity_(gravity) {
   keys_.push_back(pose_key);
 }
 
 Vector StaticWrenchFactor::unwhitenedError(
-    const Values &x, boost::optional<std::vector<Matrix> &> H) const {
+    const Values &x, gtsam::OptionalMatrixVecType H) const {
```
gchenfc, line 42: This doesn't have a default arg, but I guess people shouldn't really call this function directly anyway

gtdynamics/universal_robot/Robot.h
```diff
@@ -156,24 +156,26 @@ class Robot {
...
       const gtsam::Values &known_values, size_t t = 0,
-      const boost::optional<std::string> &prior_link_name = boost::none) const;
+      const std::optional<std::string> &prior_link_name = {}) const;
 
  private:
   /// Find root link for forward kinematics
   LinkSharedPtr findRootLink(
       const gtsam::Values &values,
-      const boost::optional<std::string> &prior_link_name) const;
+      const std::optional<std::string> &prior_link_name) const;
```
gchenfc, line 165: should this have a default argument?  or is it strict that user should explicitly specify {} if they don't want to provide it?

gtdynamics/optimizer/EqualityConstraint.cpp
```diff
@@ -16,7 +16,7 @@
 namespace gtdynamics {
 
 gtsam::NoiseModelFactor::shared_ptr DoubleExpressionEquality::createFactor(
-    const double mu, boost::optional<gtsam::Vector&> bias) const {
+    const double mu, std::optional<gtsam::Vector> bias) const {
```
gchenfc, line 19: Notice the ref went away, wondering if there's anything to gain from using `std::ref` around vector (to avoid copies) or not.

gtdynamics/utils/Trajectory.cpp
```diff
@@ -162,7 +161,10 @@ void Trajectory::writeToFile(const Robot &robot, const std::string &name,
   for (auto &&joint : robot.joints()) {
     jnames.push_back(joint->name());
   }
-  string jnames_str = boost::algorithm::join(jnames, ",");
+  string jnames_str = "";
+  for (size_t j = 0; j < jnames.size(); j++) {
+    jnames_str += jnames[j] + (j != jnames.size() - 1 ? "," : "");
+  }
```
gchenfc, line 167: I might be being dumb but why not just construct the `jnames_str` in the first for-loop instead of constructing the `jnames` vector then doing another loop to construct the string? (and pop_back the last comma)

See also: [stackoverflow `std::copy` one-liner join](https://stackoverflow.com/a/5689042/9151520)

gchenfc (approved):
LGTM

You can see a few comments I made, but I think they're all pretty insignificant.  I think the CI will be the real reviewer 😂

gchenfc:
> @gchenfc @yetongumich I haven't updated the cable robot and the jumping robot. Can you make a PR against this PR to address that?

Is that necessary for CI to pass?

gtdynamics/universal_robot/Robot.h
yetongumich: This is a helper function, which is only called in `forwardKinematics` that always pass in the optional typed variable.

Personally I think it's fine to not provide a default argument here.

gtdynamics/dynamics/Chain.cpp
```diff
@@ -121,7 +121,7 @@ gtsam::Vector3 Chain::DynamicalEquality3(
     gtsam::OptionalJacobian<3, 3> H_angles,
     gtsam::OptionalJacobian<3, 3> H_torques) {
   Matrix J;
-  poe(angles, boost::none, J);
+  poe(angles, {}, J);
```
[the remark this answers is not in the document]
varunagrawal: It should be `nullptr`. I should update this.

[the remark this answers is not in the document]
varunagrawal: Interesting, for `std::optional`, the equivalent of `boost::none` is `{}`. `nullptr` throws a compilation error.

gtdynamics/statics/StaticWrenchFactor.cpp
varunagrawal: This is the .cpp file. The .h file has the default arg so I guess I can mark this resolved.

gtdynamics/universal_robot/Joint.h
```diff
@@ -284,20 +288,20 @@ class Joint : public boost::enable_shared_from_this<Joint> {
    * Return transform of child link CoM frame w.r.t parent link CoM frame
    */
   Pose3 parentTchild(double q,
-                     gtsam::OptionalJacobian<6, 1> pMc_H_q = boost::none) const;
+                     gtsam::OptionalJacobian<6, 1> pMc_H_q = {}) const;
```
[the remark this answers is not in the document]
varunagrawal: I have no idea. I simply did a replacement based on the type without making too many semantic changes.
At least for `std::optional<gtsam::VectorN>` it's just an optional parameter and unrelated to `OptionalMatrixType`/`OptionalJacobian`.

gtdynamics/optimizer/EqualityConstraint.cpp
varunagrawal: I was wondering about that too. `std::optional` doesn't like references right out of the box and GTSAM doesn't use reference types for optionals so figured this is okay.

gtdynamics/utils/Trajectory.cpp
varunagrawal: Just tried to avoid any semantic changes. Someone else can update this if the efficiency is needed.

varunagrawal:
> > @gchenfc @yetongumich I haven't updated the cable robot and the jumping robot. Can you make a PR against this PR to address that?
> 
> Is that necessary for CI to pass?

Yes since those flags are ON. I'll turn them OFF but when you make the PRs, please enable them again.

gtdynamics/factors/CollocationFactors.h
varunagrawal: I'm following the same defaults as GTSAM for consistency.

gtdynamics/utils/Trajectory.cpp
varunagrawal: I saw the one-liner before but it adds a delimiter at the end which needs to be removed, and you need to declare an output stringstream, so it isn't quite a one-liner.

PR #371, opened 2023-02-13
Update Manifold Optimizer

varunagrawal:
This PR updates the manifold optimization PR to be compliant with GTSAM 4.3.


yetongumich (approved):
Thanks for having this done @varunagrawal !
I thought I was supposed to work on this, but you are just so fast and ahead.
Managed to get the code compiled and run in my environment, should be ready to merge.

varunagrawal:
Thanks! This should allow us to get a jumpstart on the constraint manifold state estimation paper.

PR #376, opened 2023-05-30
Fix Python CI workflow 

danbarla:
Python CI doesn't pass, since it can't install dependencies for ubuntu 22.
happened for both ubuntu 22 gcc and clang.


danbarla:
Hi @varunagrawal , Could you please approve this one-liner?

varunagrawal (approved):
LGTM

PR #378, opened 2023-06-22
DynamicsSymbol needs #include<limits> when boost serialization is OFF

arutkowski:
This PR fixes an error that is encountered when attempting to compile GTDynamics when `GTSAM_ENABLE_BOOST_SERIALIZATION` is OFF.

The error encountered with gcc-11 is 'numeric_limits' is not a member of 'std'


varunagrawal (approved):
LGTM

PR #381, opened 2023-09-18
Fix TBB linking

varunagrawal:
Update the CMake files to allow for correct linking of TBB, which is currently failing in the CI.

The main change is to delete the `FindTBB.cmake` file which allows CMake to use the `FindTBB` version provided by TBB instead.


achintyamohan (approved):
Looks good to me

PR #383, opened 2023-09-28
Use NoiseModelFactorN as factor base class

varunagrawal:
To be consistent with recent updates, upgrading the factors to use @gchenfc's new `NoiseModelFactorN` base class.


gchenfc (approved):
LGTM!

PR #385, opened 2023-10-09
Fix all the scripts

varunagrawal:
Fixed some scripts which were broken because of mine and Frank's previous PRs.

Basically, I had to initialize the prior joint angles, velocities and torques.


gchenfc (commented):
It's been a long time since I worked on this codebase, but would utils/Initializer::ZeroValues work here?

varunagrawal:
@gchenfc gentle reminder

yetongumich (approved):

PR #387, opened 2023-10-10
Minor Directory Cleanup

varunagrawal:
Moved some files to semantically better locations.


gchenfc (approved):
I assume nao.urdf is not referenced anywhere in the actual codebase so no path updates needed to happen

PR #388, opened 2023-10-19
Fix inconsistent test results

### LINKED ISSUE #233: Test for Quadruped Torque gives different values on differents OSes
The Quadruped test in `testStaticsSlice.cpp` gives different values for the torque on Linux vs MacOS.

- Linux: 0.0670426
- Mac: 0.15180275491120632

varunagrawal:
Fixes #233.

The issue was that `Statics::solve` was calling `initialValues(slice, robot)` with a default gaussian noise parameter of 0.1. This is neither a functional style and is machine dependent due to the PRNG.

I added `gaussian_noise` as an extra parameter, and set all the `gaussian_noise` defaults to be a more sensible 0.0. The user is free to update them if desired.

### DIFF
--- a/gtdynamics/kinematics/Kinematics.h
+++ b/gtdynamics/kinematics/Kinematics.h
@@ -148,12 +148,12 @@ class Kinematics : public Optimizer {
    *
    * @param context Slice or Interval instance.
    * @param robot Robot specification from URDF/SDF.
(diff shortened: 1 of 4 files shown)


varunagrawal:
@gchenfc gentle reminder

gchenfc (approved):
nice

PR #391, opened 2024-10-01
Update CMake minimum version

varunagrawal:
CMake is throwing the warning:
```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

This PR updates all the CMake files to have the minimum version as 3.5 to fix any potential deprecations as well as maintain the largest generality.


varunagrawal:
Merging

PR #418, opened 2026-02-13
CableRobot additions

dellaert:
This PR takes all the files from #331 that were just added to the repo. And then two simple commits on top of that to make it compile and test. All tests pass in C++ and Python.


gchenfc (approved):
LGTM!

PR #425, opened 2026-06-05
Docx changes

kshaji3:
-Made the installation directions updated for GCC/G++ 13 and Ubuntu 24.04
-Slightly cleaned up the README.md so there's no more reason for horizontal scrolling of commands
-Added compatibility fixes to the latest version of GTSAM's handling of NonlinearFactorGraph
-Added scheduled running of CI to make sure everything passes once a week.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,7 @@ models/mujoco/unitree_g1
 models/mujoco/boston_dynamics_spot
 /webdemo/.vite/deps/_metadata.json
 /webdemo/.vite/deps/package.json
+
+# sdformat-15.0.0
+sdformat-15.0.0
+sdformat-15.0.0.tar.bz2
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -129,3 +129,7 @@ models/mujoco/unitree_g1
 models/mujoco/boston_dynamics_spot
 /webdemo/.vite/deps/_metadata.json
 /webdemo/.vite/deps/package.json
+
+# sdformat-15.0.0
+sdformat-15.0.0
```
dellaert: You should not add your own personal install temps to .gitignore

kshaji3: I thought I had a good reason for this (if you follow the installation instructions for Linux, this is how things get installed, so this should be true for most people), but I realize that is not correct.

dellaert (approved):

PR #429, opened 2026-06-22
Adding End Effector SE(3) Targets 

kshaji3:
This PR adds the ability to specify an end effector SE(3) position target (previously it was limited to only position, so this adds orientation) with the goal of being mainly applicable to manipulation platforms (such as that used in https://www.bar-lab.org/).
- Also added a unit test for this based on BAR Lab code in the Python wrapper.


gtdynamics/kinematics/Kinematics.h
```diff
@@ -322,7 +330,8 @@ class Kinematics : public Optimizer {
   template <class CONTEXT>
   gtsam::Values inverse(
       const CONTEXT& context, const Robot& robot, const PoseGoals& pose_goals,
-      const gtsam::Values& joint_priors = gtsam::Values()) const;
+      const gtsam::Values& joint_priors = gtsam::Values(),
```
dellaert, line 333: GTSAM naming convention, already violated in this file, but let's not add to it. E.g., `poseGoals`, `enforceAsConstraints`

dellaert (approved):
LGTM. CI fails, unrelated, but you could maybe fix it as well so board is green again

gtdynamics/kinematics/Kinematics.h
kshaji3: With the help of AI, I fixed this and checked over every part of the codebase, and I believe it is now consistent with the GTSAM conventoins.

PR #431, opened 2026-07-09
GPMP2 Addition

kshaji3:
Ports over code from GPMP2 that is relevant for manipulator platforms, also updates them to the latest GTSAM standards. This also lays down some of the framework that's needed to eventually incorporate STEAP into the repo as well.
https://github.com/gtrll/gpmp2
Notes:

1. Adds the ability to define obstacle factors, including GP-based ones
2. Adds the ability to define Linear, Lie, and Pose3 GP prior factors, along with their interpolation cousins
3. Also adds joint, velocity factors
4. Enables for path planning around obstacles via GPMP
5. Includes relevant unit tests.

The code was tested in simulation and validated on real hardware (using the Python build)

### DIFF
--- a/gtdynamics.i
+++ b/gtdynamics.i
@@ -1140,16 +1140,12 @@ class ObstacleSDFFactorGP : gtsam::NoiseModelFactor {
 
 #include <gtdynamics/factors/SelfCollisionFactor.h>
 class SelfCollisionPair {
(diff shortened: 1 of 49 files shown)


kshaji3:
I have broken up this PR into several components (see other PRs). Therefore I'm closing it.

PR #432, opened 2026-07-25
Add GPMP2 core: GP interpolation and priors, joint/velocity limits

kshaji3:
Ports the Gaussian process motion planning building blocks that do not depend on obstacle or self-collision costs: which includes the all the GP-based functionality across Lie Groups and interpolation.


dellaert (approved):
LGTM!!!

PR #434, opened 2026-07-25
Gpmp2 self collision

kshaji3:
This PR adds selfCollisionFactor and selfCollisionFactorGP. This PR has contents that were tested on an 18DOF manipulator setup.
- selfCollisionFactor checks at nodes, while selfCollisionFactorGP ensures that the interpolated trajectory is collision-free. Thi

This uses selfCollisionFactor from: https://arxiv.org/pdf/2011.07171, while adding in a custom GP-implementation of it.


dellaert (approved):
Cool!

PR #437, opened 2026-07-30
Adds Sphere Radii to the ObstacleGP factors

kshaji3:
This PR adds the ability to do sphere-based obstacle avoidance in interpolation by adding it to ObstacleSDFFactorGP. It follows similarly to how this was done in selfCollisionSphereFactor.

There sphere-based obstacle avoidance already exists in ObstacleSDFFactor, so it's not needed to be added there.


dellaert (commented):
See comment above

dellaert (commented):
Three follow-ups on organizing the now-consolidated implementation:

gtdynamics/gpmp2/detail/collisionValidation.h
```diff
@@ -0,0 +1,51 @@
+/* ----------------------------------------------------------------------------
+ * GTDynamics Copyright 2020, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * See LICENSE for the license information
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file  collisionValidation.h
```
dellaert: Please use the GTSAM-style `internal` mechanism here rather than `gpmp2/detail`. Put both files under `gtdynamics/factors/internal/CollisionFactorUtils.{h,cpp}`, place the declarations and definitions in `gtdynamics::internal`, and list the `.cpp` explicitly in CMake. These functions implement factor construction, so `factors/internal` is their natural home. If the factors become header-only as suggested separately, install the `internal` header as GTSAM does with `gtsam/nonlinear/internal`: `internal` means unsupported implementation API, while still allowing installed headers to compile.

gtdynamics/gpmp2/detail/framedObstacleCost.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param wTs pose of the field's frame s in the world frame
+ * @param point query point, in the world frame
+ * @param sdf signed distance field, expressed in frame s
+ * @param epsilon standoff distance at which the cost becomes non-zero
+ * @param Hpose optional Jacobian of the cost with respect to wTs
+ * @param Hpt optional Jacobian of the cost with respect to the point
+ * @return the hinge loss cost
+ */
+double hingeLossObstacleCost(const gtsam::Pose3 &wTs,
```
dellaert: This overload was already part of the installed `ObstacleCost.h` API on `master` and was declared with `GTSAM_EXPORT`. Moving it into an uninstalled header silently breaks downstream C++ callers, and removing the export can prevent the existing test executable from linking on Windows. Please restore the declaration and documentation to `ObstacleCost.h` with `GTSAM_EXPORT`; its implementation can remain in `ObstacleCost.cpp`.

gtdynamics/factors/ObstacleSDFFactor.cpp
```diff
@@ -12,75 +12,39 @@
...
+      epsilon_(epsilon),
+      radii_(radii),
+      robot_(robot),
+      sdf_(sdf) {
+  validateObstacleSDFFactorArgs(*robot_, sdf_, epsilon_, radii_,
+                                "ObstacleSDFFactor");
 }
 
 /* ************************************************************************* */
 gtsam::Vector ObstacleSDFFactor::evaluateError(
```
dellaert: The four factor `.cpp` files are now just delegating constructors plus one-line `evaluateError` methods. I recommend making `ObstacleSDFFactor`, `ObstacleSDFFactorGP`, `SelfCollisionSphereFactor`, and `SelfCollisionSphereFactorGP` header-only again: move these definitions into their public headers and remove the four tiny `.cpp` files, while keeping the actual collision calculations in the shared helpers. This works with the installed `internal` header suggested above; remember that any out-of-line helper symbols called by installed inline definitions must be exported on Windows.

gtdynamics/gpmp2/detail/collisionValidation.h
kshaji3: Ok, I saw this existed, but was nowhere in GT Dynamics. I will follow this convention

gtdynamics/factors/ObstacleSDFFactor.cpp
kshaji3: I believe latest edits address this properly.

gtdynamics/gpmp2/detail/collisionValidation.h
kshaji3: Fixed now I believe.

gtdynamics/gpmp2/detail/framedObstacleCost.h
kshaji3: Fixed now I believe.

dellaert (approved):
Looks great now! Thanks for being so responsive!

